Issue #33, opened 2022-10-28
Brand work for website


AnthonyRollins:
## To-Dos
- [x] Create Figma components for existing website designs.
- [x] Update styling (font size, colors, spacing) for all webpages
- [ ] Produce mobile mockups of pre-existing pages.

segacy1:
Hi @AnthonyRollins! Do we have brand guidelines at the moment? I was poking around in drive and came across [this](https://drive.google.com/drive/folders/1_tXVp94IS5UKMIP8pHJOmarmaYCqrCWx?usp=sharing), but not sure if these docs are current. Want to make sure I'm using our styles as intended/document essentials in figma

AnthonyRollins:
@segacy1 Yes, that is the right folder. I've been using this [style guide](https://drive.google.com/drive/folders/1GkzWwdCzbdpXKnxAeBya6tWYJWPFEZ5i), as its the most up-to-date.

segacy1:
@AnthonyRollins @thekaveman @angela-tran hey folks! I’m making some progress on the UI library and I wanted to get some feedback on a couple of things before I kept going. The section I'd like feedback on is labeled "Dark Mode Update" [here](https://www.figma.com/file/2gq3WJFdRFDefyHcgFMUGt/Current-compiler.la?node-id=313%3A7116).

There are a few additions that we don't currently use on the live site, but I could foresee them being good to have as we build out the site further. Keep in mind, everything’s still a WIP!


**Color Naming Conventions**

_What I did:_ 
I segmented our color system into a few sections: Actions, Background, Text, and Brand. For dev, I assumed that it was easier for colors to be named by function. For design, I think having colors labeled by name is easier as we might create things that aren’t necessarily tied to function. That is what the brand section is for.

_Question:_
Does this work for you all? Anything that could be improved for you here?

**Light Mode vs. Dark Mode**

_Questions:_
- The current styling is all on a black background. Would we want to use only black backgrounds moving forward? Do we want an option for light mode/backgrounds? 
- The color palette that was designed for compiler was designed for a white background. I updated the two secondary colors (purple and indigo) to work better on a black background. Please let me know if that doesn't feel right - I can keep finessing the hues if so.

**Buttons + Component Anatomy**

_What I did:_ 
- To keep our buttons scalable for screen sizes and new designs, I created 3 sizes: large, normal, and small. 
- I added additional states for our two button types (primary/secondary). 

 _Questions:_
[...]

PR #43, opened 2023-02-09
Refactor: use Bootstrap

### LINKED ISSUE #36: Refactor layouts to use Bootstrap
As we move towards adding more information to our website and towards restyling some parts (e.g. #24 and #33), it would be beneficial to use Bootstrap rather than writing all the CSS ourselves.

This refactor involves [adding Bootstrap](https://getbootstrap.com/docs/5.2/getting-started/download/#cdn-via-jsdelivr), making necessary changes to preserve the existing visual layout/styles, and then cleaning up unused CSS from our stylesheet.

angela-tran:
Closes #36 

This PR refactors our layouts to use Bootstrap and tries to maintain the same net effect in terms of styling and spacing.

Note that the attempt to maintain styling/spacing is approximate due to time and lack of necessity -- this means there are some minor discrepancies between what's live on compiler.la vs. this branch. Since we'll be implementing more re-designs (#33), I think this is acceptable. The biggest discrepancy is just on the responsive behavior of the homepage buttons (which I think is ok):

image

This should be ok to merge in whenever approved -- let me know if you notice any unexpected regression!

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -58,8 +58,8 @@
             {% include certs.html %}
 
         </div>
-        <div class="">
-            <div class="">
+        <div class="d-flex social-media-container">
+            <div class="d-flex justify-content-between align-items-end py-4">
                 <a href="https://github.com/compilerla"><img class="social-media-icon"
                         src="/assets/social_media/github_mark.png" alt="GitHub"></a>
                 <a href="https://twitter.com/CompilerLA"><img class="social-media-icon"
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -46,8 +46,8 @@
     </main>
 
     <footer>
-        <div class="divider-line my-5"></div>
-        <div class="certs">
+        <div class="divider-line"></div>
+        <div class="certs d-flex flex-column flex-lg-row row-cols-lg-4">
             <div>
                 <a href="mailto:hello@compiler.la" class="monospace contact-link">hello@compiler.la</a>
                 <p class="address">
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -44,7 +44,7 @@
 
     <footer>
         <div class="divider-line"></div>
-        <div class="row certs shifted-down">
+        <div class="certs">
             <div>
                 <a href="mailto:hello@compiler.la" class="monospace contact-link">hello@compiler.la</a>
                 <p class="address">
@@ -55,17 +55,14 @@
             {% include certs.html %}
 
         </div>
-        <div class="row">
-            <div class="social-media-container">
-                <a href="https://github.com/compilerla">
-                    <img class="social-media-icon" src="/assets/social_media/github_mark.png" alt="GitHub">
-                </a>
-                <a href="https://twitter.com/CompilerLA">
-                    <img class="social-media-icon" src="/assets/social_media/twitter_logo.svg" alt="Twitter">
-                </a>
-                <a href="https://www.linkedin.com/company/compiler-la">
-                    <img class="social-media-icon" src="/assets/social_media/linkedin_in.png" alt="LinkedIn">
-                </a>
+        <div class="">
+            <div class="">
+                <a href="https://github.com/compilerla"><img class="social-media-icon"
+                        src="/assets/social_media/github_mark.png" alt="GitHub"></a>
+                <a href="https://twitter.com/CompilerLA"><img class="social-media-icon"
+                        src="/assets/social_media/twitter_logo.svg" alt="Twitter"></a>
+                <a href="https://www.linkedin.com/company/compiler-la"><img class="social-media-icon"
+                        src="/assets/social_media/linkedin_in.png" alt="LinkedIn"></a>
             </div>
             <img class="brandmark" src="/assets/compiler_brandmark.svg" alt="Compiler brandmark">
         </div>
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -20,6 +20,9 @@
     <link rel="icon" href="/icon.svg" type="image/svg+xml">
     <link rel="apple-touch-icon" href="/apple-touch-icon.png">
 
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-rbsA2VBKQhggwzxH7pPCaAqO46MgnOM80zW1RWuH61DGLwZJEdK2Kadq2F9CUG65" crossorigin="anonymous">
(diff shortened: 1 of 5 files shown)


angela-tran:
Putting this in draft. I will try to get some of the font-sizes back to what they are on the live site.

styles/base.css
```diff
@@ -42,40 +54,106 @@ a {
...
+a:hover,
+a:focus {
     color: var(--green);
 }
 
 ul.no-bullets {
+    padding-left: 0;
+}
+
+ul.no-bullets li {
```
machikoyasuda, line 66: Could you use this Bootstrap unstyled list instead: https://getbootstrap.com/docs/5.2/content/typography/#unstyled

_layouts/default.html
```diff
@@ -20,6 +20,9 @@
     <link rel="icon" href="/icon.svg" type="image/svg+xml">
     <link rel="apple-touch-icon" href="/apple-touch-icon.png">
 
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-rbsA2VBKQhggwzxH7pPCaAqO46MgnOM80zW1RWuH61DGLwZJEdK2Kadq2F9CUG65" crossorigin="anonymous">
+    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/js/bootstrap.bundle.min.js" integrity="sha384-kenU1KFdBIe4zVF0s0G1M5b4hcpxyD9F7jL+jjXkk+Q2h455rYXK/7HAuoJl+0I4" crossorigin="anonymous"></script>
```
machikoyasuda, line 24: Do we need the JS (yet)? If we are not using the JS, we can remove this for now. 

What uses Bootstrap JS:
image

styles/base.css
```diff
@@ -42,40 +54,106 @@ a {
...
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
     font-family: 'Source Code Pro Bold', 'Courier New', Courier, monospace;
 }
 
 .san-serif {
```
machikoyasuda, line 94: `.sans-serif` (Sans means without in French.) https://en.wikipedia.org/wiki/Sans-serif

styles/base.css
```diff
@@ -42,40 +54,106 @@ a {
...
+
+.monospace,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
     font-family: 'Source Code Pro Bold', 'Courier New', Courier, monospace;
 }
```
machikoyasuda, line 92: Looking at the Style Guide, https://www.figma.com/file/2gq3WJFdRFDefyHcgFMUGt/Current-compiler.la?node-id=326-322&t=qbG8SaSrYZENtOYA-0, only H1 and H2 are Source Code Pro.

styles/base.css
```diff
@@ -106,39 +184,66 @@ li.certs {
...
 .logo {
-    width: 20rem;
+    width: var(--logo-width);
 }
 
+/* Main */
+
 main {
     flex: 1 0 auto;
 }
```
machikoyasuda, line 214: Remove

styles/base.css
```diff
@@ -106,39 +184,66 @@ li.certs {
...
+@media (min-width: 992px) {
+    .divider-line {
+        margin: 20px -30px 20px;
+    }
+
 }
 
 footer {
     flex-shrink: 0;
 }
```
machikoyasuda, line 233: Remove

styles/base.css
```diff
@@ -157,110 +262,37 @@ footer .address {
...
+
+@media (min-width: 992px) {
+    .social-media-icon {
+        margin: 0 24px;
+        width: 3rem;
+    }
 }
 
 .social-media-container a:first-child .social-media-icon {
     margin-left: 0px;
```
machikoyasuda, line 280: 0 doesn't need `px` after

machikoyasuda (changes requested):

styles/base.css
machikoyasuda: Actually - Just checked the rest of the repo. The repo only uses h1 and h2 so far, so let's just remove h3-h6 here.

styles/base.css
```diff
@@ -42,40 +54,106 @@ a {
...
-.shifted-down {
-    margin-top: 50px;
+h5 {
+    font-size: 0.83rem;
 }
 
+h6 {
+    font-size: 0.67rem;
+}
+
```
machikoyasuda, line 154: I think(?!?) these h3-h6 styles can be removed cos the app is not using any yet.

jobs.html
```diff
@@ -4,7 +4,7 @@
 ---
 <h1>Jobs</h1>
 
-<h2>Accepting applications</h2>
+<h2">Accepting applications</h2>
```
machikoyasuda, line 7: Remove `"`

_layouts/default.html
```diff
@@ -44,7 +47,7 @@
 
     <footer>
```
machikoyasuda, line 48: I _think_ instead of adding `padding` to `<body>`, `<main>` should have `class="container"`, which will give it responsive max-widths. All `rows` need to be in a `container`.

_layouts/default.html
```diff
@@ -44,7 +47,7 @@
...
         <div class="divider-line"></div>
-        <div class="row certs shifted-down">
+        <div class="certs d-flex flex-column flex-lg-row row-cols-lg-4">
```
machikoyasuda, line 50: I actually don't think any of these flex classes are necessary. If this was a `div` with a `row` class, with four child `div`s with a `col-md-3`, then the columns would respond the way we want them to: 100% width in small widths, and then 25% on wider screens.

styles/base.css
machikoyasuda: Save this for a future refactor.

styles/base.css
machikoyasuda: Save this for a future refactor. Resolving for now.

_layouts/default.html
machikoyasuda: Save this for a future refactor. Resolving for now.

styles/base.css
```diff
@@ -15,24 +15,36 @@
...
+@media (min-width: 992px) {
+    :root {
+        --body-padding: 30px;
+    }
+}
+
 body {
     display: flex;
     flex-direction: column;
```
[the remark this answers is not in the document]
machikoyasuda: Save this for a future refactor. Resolving for now.

styles/base.css
machikoyasuda: Save this for a future refactor. Resolving for now.

machikoyasuda:
After discussing with Angela over Slack, I've decided and let her know that I'm resolving most comments above as ("Save this for a future refactor. Resolving for now."). The only actual fix I want to see is the HTML error (h2"). Otherwise, this looks good to me. I'd like to see this PR merged, and then I can start reviewing #44

thekaveman (approved):

PR #46, opened 2023-02-27
update product manager role, re-open

vykster:
@angela-tran I think it's easy to miss the "Apply Now" link. Would it be reasonable to style this as a button?


angela-tran:
@vykster How does the link look now?

angela-tran (approved):
Feel free to merge in if the link button-style looks good to you @vykster ! 💯

PR #53, opened 2023-03-20
Fix: Apply Now button

### LINKED ISSUE #52: Text appears on top of button on job posting
## Context
The bottom content appears over the Apply Now button on the latest job posting for the [Senior Consultant, Product Manager](https://compiler.la/jobs/product-manager). This position no longer accepts applications, so this isn't an urgent fix, but it should be resolved before our next post. No other job posting pages are affected as they don't have the Apply now button.

## Screenshots
Desktop
Screenshot 2023-03-20 at 3 23 06 PM

Mobile
Screenshot 2023-03-20 at 3 22 36 PM

angela-tran:
Fixes #52 

This fixes the issue as well as extracts the link out to an include file that can accept a URL.

### DIFF
--- a/_includes/job-apply-button.html
+++ b/_includes/job-apply-button.html
@@ -1,0 +1,1 @@
+<a class="d-inline-block monospace text-lg primary-btn" href="{{ include.link }}">Apply Now</a>
--- a/_jobs/executive-assistant.md
+++ b/_jobs/executive-assistant.md
@@ -75,7 +75,7 @@ This is a full-time, salaried position (W2) with benefits. Full-time employment 
 * Cover letters are not expected and extra weight in scoring will not be given to those that include them
 * Writing samples will be requested as part of second round interviews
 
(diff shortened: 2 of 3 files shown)


thekaveman (commented):
I like this include implementation. I wonder if it makes sense to conditionally show the apply button based on whether the job is active or not?

Some of the previous postings we just had a "send an email" link, it was maybe easier to ignore after being closed. But I worry with a giant green button, people will still click it even after closed, and I think the ClickUp form would still accept?

machikoyasuda:
@thekaveman @angela-tran Maybe for jobs where the application window has passed, the button could be disabled? with the `disable` https://getbootstrap.com/docs/5.3/components/buttons/#disabled-state

angela-tran:
Here's what the disabled state looks like when I add `btn disabled` to the list of classes. I also added `role="button"` and `aria-disabled="true"` as specified in that Bootstrap page you linked @machikoyasuda.

PR #81, opened 2023-06-16
Feat: explicit open/close dates for jobs

### LINKED ISSUE #54: Refactor job `open` boolean to instead be an expiration date
Currently our job page [looks for a variable named `open`](https://github.com/compilerla/compiler.la/blob/0ee42f8fd4078663cc94def80933d9468bc4c131/jobs.html#L8) for each job to decide whether it shows up in "Accepting applications" or "Past opportunities". This means that when a job is no longer open, we have to open a pull request to update the boolean. (See #42 and #48.)

We could instead have an expiration date variable on the job and have logic that checks the relationship between that date and the current date.

I'm not sure what the implementation would look like. This would have to be dynamic logic on the page rather than generated by Jekyll.

thekaveman:
Closes #54 

Closes https://app.clickup.com/t/863gyj9pg

## Confirming date logic works

Running the preview site from UTC+4 shows the correct `America/Los_Angeles` time:

image

## Confirming jobs display in the correct list

Added a test job with a current open window:

image

And confirmed it showed up under the "Accepting applications" heading:

image

## Confirming Apply link for open/closed jobs

The test open job:

image

A closed job:

image

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,24 +1,23 @@
 {
-    "name": "compiler.la",
-    "dockerComposeFile": "docker-compose.yaml",
-    "service": "site",
-	"workspaceFolder": "/srv/jekyll",
-    "postAttachCommand": [
-        "/bin/bash",
-        ".devcontainer/postAttach.sh"
-      ],
-    "settings": {
-        "terminal.integrated.defaultProfile.linux": "bash",
-        "terminal.integrated.profiles.linux": {
-            "bash": {
-                "path": "/bin/bash"
-            }
-        }
-    },
-    "extensions": [
-		"eamodio.gitlens",
-		"mhutchie.git-graph",
-		"redhat.vscode-xml",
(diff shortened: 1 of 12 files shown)


angela-tran (dismissed):
This looks great @thekaveman! Should we create a separate ticket for updating the Jobs Apply button template? It currently [uses the `open` variable](https://github.com/compilerla/compiler.la/blob/main/_includes/job-apply-button.html#L1) to show an enabled button.

thekaveman:
> This looks great @thekaveman! Should we create a separate ticket for updating the Jobs Apply button template? It currently [uses the `open` variable](https://github.com/compilerla/compiler.la/blob/main/_includes/job-apply-button.html#L1) to show an enabled button.

Thanks for this call-out. I think we should probably update it here, as otherwise it will be broken for new jobs! Will do a follow-up commit (or more).

angela-tran (approved):
This looks great @thekaveman !

Issue #65, opened 2023-04-21
Development work for Home page


AnthonyRollins:
This is a placeholder for the development work that will take place this quarter. Issues will be built out once designs and content are done.

AnthonyRollins:
Here's the final copy/design for the new [home page](https://www.figma.com/file/2gq3WJFdRFDefyHcgFMUGt/Current-compiler.la?type=design&node-id=1355-3984).

machikoyasuda:
## Components
- Header (h1)
- Paragraph (p)
- Button (primary-button)
- Nav height

PR #87, opened 2023-07-03
Refactor: jobs page styling

### LINKED ISSUE #82: Update Jobs page styling
Segacy has updated the styling for the jobs page. You can find the styling in [Figma](https://figma.com/file/2gq3WJFdRFDefyHcgFMUGt/Current-compiler.la?type=design&node-id=677-1890&t=yWGVmWiKAu2G6aBd-0). 

Let me know what other info you need.

angela-tran:
Closes #82

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -1,3 +1,6 @@
+---
+large_logo: false
+---
 <!DOCTYPE html>
 <html lang="en">
 
@@ -38,7 +41,12 @@
 
 <body>
     <header>
-        <a href="/"><img class="logo" src="/assets/compiler_logo_stacked.svg" alt="Compiler logo"></a>
+        {% if page.large_logo %}
+            {% assign logo_class = "logo-lg" %}
+        {% else %}
+            {% assign logo_class = "logo" %}
+        {% endif %}
+        <a href="/"><img class="{{ logo_class }}" src="/assets/compiler_logo_stacked.svg" alt="Compiler logo"></a>
     </header>
 
     <main>
--- a/jobs.html
+++ b/jobs.html
@@ -9,12 +9,12 @@ title: Jobs with Compiler
                 accessible for all.</h1>
 
             <h2 class="mt-5">Open roles</h2>
-            <ul id="open" class="mx-3"></ul>
+            <ul id="open" class="no-bullets text-decoration-none"></ul>
 
             <h2 class="mt-5">Past roles</h2>
-            <ul id="closed" class="mx-3"></ul>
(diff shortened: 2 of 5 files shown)


jobs.html
```diff
@@ -2,18 +2,28 @@
...
 
-<h2>Accepting applications</h2>
-<ul id="open" class="mx-3"></ul>
+            <h2 class="mt-5">Open roles</h2>
+            <ul id="open" class="no-bullets text-decoration-none"></ul>
 
-<h2>Past opportunities</h2>
-<ul id="closed" class="mx-3"></ul>
+            <h2 class="mt-5">Past roles</h2>
+            <ul id="closed" class="no-bullets"></ul>
```
machikoyasuda, line 16: You can use Bootstrap's `list-unstyled` here instead of `no-bullets`: https://getbootstrap.com/docs/5.2/content/typography/#unstyled

angela-tran: Done in [122bbe2](https://github.com/compilerla/compiler.la/pull/87/commits/122bbe2a6ba7a6c002daaac5db6280507a75b2b5)

_layouts/default.html
```diff
@@ -1,3 +1,6 @@
+---
+large_logo: false
```
angela-tran, line 2: This is needed so that this PR doesn't change the Homepage. We can remove this when implementing #65.

jobs.html
```diff
@@ -2,18 +2,28 @@
 layout: default
 title: Jobs with Compiler
 ---
-<h1>Jobs</h1>
+<div class="jobs container mt-5">
+    <div class="row">
+        <div class="col-md-8 offset-md-2 mt-5">
+            <span class="label rounded p-2">Careers</span>
```
thekaveman, line 8: I think something like this is normally called a `pill`

angela-tran: Renamed it in [6036943](https://github.com/compilerla/compiler.la/pull/87/commits/6036943f82145b085d51128f17206804e54f3e61)

machikoyasuda (approved):

machikoyasuda:
This can be changed in future PR's b/c it's not a big deal - but I'm not sure if this is actually rendering font-weight 600 (Semi Bold) font: I typed in `CAREERS` here at weight 600, size 12px here (https://fonts.google.com/specimen/Source+Code+Pro?preview.text=CAREERS&preview.text_type=custom) and compared it with the Netlify. I think we need to add the SemiBold font file in this directory https://github.com/compilerla/compiler.la/tree/main/fonts and make sure the SemiBold file is getting rendered.

image

PR #88, opened 2023-07-03
Refactor: improvements to job title formatting

angela-tran:
Part of #82 

Introduces a `type` variable to job post front matter. This is so job titles appear in a more consistent format, which is shown in the Figma screens for #82.

This PR also sorts jobs by title.


thekaveman (approved):
Nice improvement!

Issue #38, opened 2022-12-21
Consider max width for site


thekaveman:
Just a thought: I have a super widescreen, and the site currently expands to fill the entire viewport. Not necessarily a problem, but it might be more readable if we imposed a max width with the content center aligned and increased margins/padding on the sides.

Here's a screenshot of the homepage at full width on my screen:

<img width="1874" alt="image" src="https://user-images.githubusercontent.com/1783439/208964325-493bd703-6b14-40b3-af1f-d567c402f626.png">

And another of the current job posting:

<img width="1861" alt="image" src="https://user-images.githubusercontent.com/1783439/208964428-fa17f1b3-aa4f-41a5-a00d-c52cd16c71b6.png">

machikoyasuda:
This is done with #106 as well

PR #31, opened 2022-09-30
Feat: footer

angela-tran:
Closes #6 

Currently based off #30. Waiting for that to be merged first.

 Full page screenshot:   

image

### Proposals/questions for reviewers:
There are a few small differences between what I currently have vs. what's in the Figma.
 - **Contact section**: The Figma doesn't have a header for this section, but while working on it, I found that it looks better with a header.
    - **_Proposal_**: the Contact section has a header that says "Contact us"
 - **`hello@compiler.la` link**: The Figma shows this as a green link with a font of Source Code Pro. I think we did it like this because we were thinking of the original big-title link from the first version of the site. However, as I worked on the new link, I realized that all the other links on the site are white with Roboto as the font. The original big-title link was only styled with Source Code Pro because we were emulating the Compiler business card. Also, apparently the original link was white: https://deploy-preview-5--compilerla.netlify.app/
    - **_Proposal_**: the `hello@compiler.la` link is styled like all our other links on the site
 - **Section font**: The Figma shows the cert values as bold, white text at 18px. I'm not sure / don't remember our reasoning for this style. When I simply add the text for the cert values, they show up as normal, white text at 16px since this is the body style. This looks ok to me.
    - **_Proposal_**: no special styling for the cert values - they are styled the same as normal text on our site
 - **Social media icons**: I actually didn't notice this until I wrote out the notes above, but in our Figma, we added some space between the social media icons. I haven't made that change yet. Do y'all think this change is necessary?
    
 Let me know if you have any counter-proposals!

The original Figma mockup, from https://github.com/compilerla/compiler.la/issues/6#issuecomment-1256555734:   

image

PR #106, opened 2023-08-21
Footer

machikoyasuda:
closes #31 
closes https://github.com/compilerla/compiler.la/issues/38

## What this PR does
- **Footer**: Change fonts and spacing as necessary for new design
- **Header**: Standardizing the header margins matters for the Footer, b/c the Footer and Header must be left-aligned together. This PR uploads a new compiler logo, with flush edges, so we don't have to use negative margin CSS to remove the random space between the logo and the edge of the image.
- **CSS**: Starts to incorporate CSS Variables and Bootstrap variable overrides for Footer and Header changes.
- **Container:** All pages have `container-fluid` so we don't have this wildly wide page issue https://github.com/compilerla/compiler.la/issues/38
- **Container**: This PR removes a lot of the alignment/flex styling that was put on `body`. Instead, it uses a `--bs-gutter-x` override to declare that on Desktop, the gutter of the main `container-fluid` should be 24px on each side, and 16px on each side for Mobile, to go along with what is on Figma. As you can see from the Figma screenshots below, the x gutter should be 24px on Desktop and 16px on mobile. This change is declared here: https://github.com/compilerla/compiler.la/pull/106/files#diff-dd51c5cd06799351fdf26f10840213ef27390c39ed033c1a7d82ac493f8eb696R26-R34

image
image

## Screenshots

Figma vs. this PR:
image
image
image
image

Ensure no regressions:
image
image


machikoyasuda:
@angela-tran @thekaveman This is ready for review. Note that this is for the Footer _only_. Home Page fix PR is up next, so we won't be in limbo purgatory for too long with the footer looking different from the home page.

https://www.figma.com/file/2gq3WJFdRFDefyHcgFMUGt/Current-compiler.la?node-id=1355%3A3984&mode=dev *Just* the lower half of this page, the top of this page is to look just like compiler.la production

machikoyasuda:
Thanks to @segacy1 for sending a better tightly-cropped SVG logo!

machikoyasuda:
Ready for review @angela-tran

styles/base.css
```diff
@@ -204,22 +176,15 @@ h6 {
...
-        --header-margin-left: -10px;
         --logo-lg-width: 20rem;
     }
 }
 
-header {
-    margin-left: var(--header-margin-left);
-    /* account for padding within the Compiler logo*/
-}
-
```
machikoyasuda, line 222: This is no longer necessary, b/c the logo file is updated to be tightly cropped.

styles/base.css
```diff
@@ -8,49 +8,39 @@
...
-    flex-direction: column;
-
-    padding: var(--body-padding);
-    background-color: var(--brand-primary-black);
-
-    color: var(--white);
-    font-family: 'Roboto', Arial, Helvetica, sans-serif;
-    font-weight: normal;
-}
-
```
machikoyasuda, line 44: Using the `--bs-gutter-x` to set body padding. So this is no longer necessary. This is how the override looks like in the CSS:

image

The default `--bs-gutter-x`(left and right) is 1.5rem, but I'm setting it to 3rem/2rem (48px total/32px total responsively across Desktop/Mobile). 

Here's the 48px gutter on Desktop in effect. There is 24px on each side:
image

styles/base.css
```diff
@@ -63,40 +53,22 @@ a:focus {
...
-.jobs li {
-    margin: 1rem 0;
-}
-
-@media (min-width: 992px) {
-    .jobs li {
-        margin: 0.5rem 0 0;
-    }
-}
-
```
machikoyasuda, line 85: This is all managed with Bootstrap classes in `cert.html` instead.

styles/base.css
```diff
@@ -228,92 +193,32 @@ header {
...
-footer .certs>div {
-    margin: 30px 0 0;
-    padding-left: var(--certs-div-padding-left);
-}
-
-footer .heading {
-    font-size: 1.5rem;
```
machikoyasuda, line 271: All managed with Bootstrap utility classes in the template now.

_layouts/default.html
```diff
@@ -49,31 +49,32 @@
...
-    <footer>
-        <div class="divider-line"></div>
-        <div class="certs d-flex flex-column flex-lg-row row-cols-lg-4">
-            <div>
-                <a href="mailto:hello@compiler.la" class="monospace contact-link">hello@compiler.la</a>
-                <p class="address">
-                    525 S Hewitt St<br />Los Angeles, CA 90013
-                </p>
-            </div>
+    <footer class="border-top border-light">
```
machikoyasuda, line 56: The way this `border-light` Bootstrap class is getting its color is from the CSS root variable declaration here:

```css
    --bs-light-rgb: 84, 84, 84; /* for the HR above Footer */
    --bs-light: #545454; /* brand/scales/gray/4 */
```

More on `border-X` classes (like `border-primary`, `border-secondary`): https://getbootstrap.com/docs/5.2/utilities/borders/#color 
This also gives us the class: `text-light`: https://getbootstrap.com/docs/5.2/utilities/colors/#colors

_layouts/default.html
```diff
@@ -49,31 +49,32 @@
...
+        <div class="container-fluid">
+            <div class="row d-flex flex-column flex-lg-row row-cols-lg-4">
+                <div class="pt-lg-5 mt-lg-1 pt-4 mt-3 pb-5">
+                    <a href="mailto:hello@compiler.la" class="monospace text-primary contact-link">hello@compiler.la</a>
```
machikoyasuda, line 60: This `text-primary` class gets its color from this CSS Variable declaration:

```css
    --bs-primary: #88B440; /* Use with .text-primary */
    --bs-primary-rgb: 136, 180, 64;
```

Can also be used with `border-primary`.

styles/base.css
```diff
@@ -228,92 +194,30 @@ header {
...
     transition: 250ms;
-    margin: 0 12px;
-    width: calc(28rem / 16);
+    width: calc(24rem / 16);
 }
 
 @media (min-width: 992px) {
     .social-media-icon {
-        margin: 0 24px;
-        width: 3rem;
```
machikoyasuda, line 296: For the social media icon margins, I used the Bootstrap `gap` class instead:

`<div class="d-flex py-4 gap-md-4 gap-3">`

`gap-3` sets the children apart by 16px, and `gap-md-4` sets the children apart by 24px at medium/large screen sizes. Adding `d-flex` allows you to use gap classes.

More on gap: https://getbootstrap.com/docs/5.3/utilities/spacing/#gap

_layouts/default.html
```diff
@@ -49,31 +49,32 @@
...
-        <div class="d-flex social-media-container">
-            <div class="d-flex justify-content-between align-items-end py-4">
-                <a href="https://github.com/compilerla"><img class="social-media-icon"
-                        src="/assets/social_media/github_mark.png" alt="GitHub"></a>
-                <a href="https://www.linkedin.com/company/compiler-la"><img class="social-media-icon"
-                        src="/assets/social_media/linkedin_in.png" alt="LinkedIn"></a>
             </div>
-            <img class="brandmark" src="/assets/compiler_brandmark.svg" alt="Compiler brandmark">
+            <div class="d-flex social-media-container">
+                <div class="d-flex py-4 gap-md-4 gap-3">
```
machikoyasuda, line 70: See note about `gap` below.

styles/base.css
```diff
@@ -1,56 +1,47 @@
...
 
 :root {
-    --purple: #663190;
     --green: #88B440;
-    --teal: #01B4CB;
-    --navy: #223061;
-    --brand-primary-black: #1C1C1C;
-    --black: #000000;
     --white: #FFFFFF;
+    --brand-primary-cyan: #51B1C7;
```
machikoyasuda, line 6: `teal` is now called `cyan`, and the color changed from `#01B4CB` to `#51B1C7`

The color is called `cyan` on Figma, so I'm changing it to `cyan` here as well.

machikoyasuda:
@angela-tran Okay, _now_, it's finally ready for review. I added informational comments on various files and lines.

styles/base.css
machikoyasuda: This `bs-gutter-x` variable is used not just here, but in any class concerning container and column widths: 
image

So it's pretty cool that with CSS Variables, we can now make cascading variable changes across the app without Sass.

styles/base.css
```diff
@@ -228,92 +194,30 @@ header {
...
-    border-top: 1px solid #545454;
-    margin: 10px -16px 10px;
-}
-
-@media (min-width: 992px) {
-    .divider-line {
-        margin: 20px -30px 20px;
-    }
-
-}
```
machikoyasuda, line 249: This is now `border-light border-top` using `bs-light-rgb` color. See comment above.

machikoyasuda:
@thekaveman Hrm.. but `container` would make the header logo go too far in:

This is what `container` looks like for me:

image

and we want the logo to be aligned vertically with the Footer like this:
image

thekaveman:
You're looking at a different branch I think? It looks like the homepage updates are there. 

The issue is with `container-fluid` everywhere, it expands to the whole width:

> `.container-fluid`, which is `width: 100%` at all breakpoints

https://getbootstrap.com/docs/5.2/layout/containers/

All 3 places it was added in this PR should be `container` I think?

machikoyasuda:
@thekaveman But `container` would make the whole app show up with way too much margin left and right on 13 inch laptop size screens:

image

thekaveman:
@machikoyasuda isn't that controlled by the Bootstrap gutter?

This doesn't look problematic to me? 1024px with `container` shows the standard 32px of margin on each side of all 3 elements

image

styles/base.css
```diff
@@ -33,6 +32,12 @@ body {
     }
 }
 
+@media (min-width: 1920px) {
+    .container-fluid {
+        max-width: 1320px;
+    }
+}
```
machikoyasuda, line 39: I got the 1920px width breakpoint suggestion from CSS-Tricks: https://css-tricks.com/optimizing-large-scale-displays/#aa-build-with-relative-units They have breakpoints at 1920px, 2560px. Bootstrap breakpoints end with the largest at 1400px.

I got the 1320px width from Bootstrap: https://getbootstrap.com/docs/5.2/layout/containers/ It's the max-width for containers of screens up to 1400px.

machikoyasuda: @thekaveman Can you try this on your machine? Does 1320px feel like a good width, or should it be wider?

thekaveman:
Yeah this looks a lot better, thank you! 🚀 

image

thekaveman:
Side note: I think we could use a little bit more top spacing? Like another 0.5rem - 1rem or something

machikoyasuda:
@thekaveman Yea but that will all be changed in #107 so we could just wait for that. Or I could add a class there for more spacing that would be on prod for a day or something.

thekaveman:
> Yea but that will all be changed in https://github.com/compilerla/compiler.la/pull/107 so we could just wait for that.

All good

thekaveman (approved):

Issue #66, opened 2023-04-21
Development work for About page


AnthonyRollins:
This is a placeholder for the development work that will take place this quarter. Issues will be built out once designs and content are done.

segacy1:
Dropping assets for this page here
[about page assets.zip](https://github.com/compilerla/compiler.la/files/12403243/about.page.assets.zip)

AnthonyRollins:
@segacy1 These assets don't include Shelby. I added a comment to the About page designs regarding the same issue.

segacy1:
Added Shelby in the designs and including her image here: 
![shelby](https://github.com/compilerla/compiler.la/assets/33331853/59f8e785-3f2a-4eef-974d-262b01282229)

machikoyasuda:
@segacy1 Can you re-export the image for Shelby so it matches the image sizes of the rest of the illustrations? It's a bit short:

<img width="1319" alt="image" src="https://github.com/compilerla/compiler.la/assets/3673236/0572fc80-2e04-48f1-9a94-41a5b81c3802">

segacy1:
![shelby](https://github.com/compilerla/compiler.la/assets/33331853/aaaf2ffe-f5e1-4336-96eb-8879715bec5d)
@machikoyasuda try this one

PR #44, opened 2023-02-11
Feat: nav menu

angela-tran:
Part of #24 

Builds off #43

Issue #37, opened 2022-12-20
Update fonts and colors


angela-tran:
Once the styling updates in #33 have been finalized in Figma, we can start incorporating those changes by updating the font sizes and colors.

machikoyasuda:
<img width="822" alt="image" src="https://github.com/compilerla/compiler.la/assets/3673236/4b1b94ae-60c3-4071-af98-bd91b7ee1dfe">
<img width="861" alt="image" src="https://github.com/compilerla/compiler.la/assets/3673236/8b3a682b-019f-4fdb-abcb-e7d2c5c85da0">

machikoyasuda:
Also, make small tweaks to Button to match designs

PR #107, opened 2023-08-21
Home

### LINKED ISSUE #65: Development work for Home page

machikoyasuda:
fix #65 

Note: This PR was created off of `fix/footer` #106. These 2 changes can go out together in sync, so the Home page doesn't look disjointed.

## What this PR does
**- Home page:** Adds new home headline, paragraph text, image. Upload new brandmark. New cyan/teal color.
**- Header:** Removes the large logo logic. Add the box-shadow.
**- Whole app:** Set correct background color

## Components
- Configure `btn-primary`, `btn-outline-primary` with `:hover`, `:focus`, `:disabled` modes all designed.
- Configure `.h2` `h2` and `.h3`. The home page has a `h1` that uses a `.h2` style class. The job posting page uses a `.h3` class style on its `h2`s. This PR does *not* re-do all the headline font sizes. That will come in a separate PR #116 
- Sets focus ring for buttons and links.
- Set `text-info` to be gray color.

### Home
image
image
image

### Jobs (Ensure no regressions)

image

### Posting (Ensure no regressions)
image

image
Has the new primary disabled button

## Perf / A11y
image

image


machikoyasuda:
Outstanding question:

@thekaveman The design calls for removing the Capabilities Statement button. But since we don't have the #44 yet, we don't have any other way for people to get to it. Should I keep the button there? Should I remove the Get in touch button, so at least the design is the same with 1 button? There is a contact email link right below it.

machikoyasuda:
Marking this back as `In Progress` as I update the outline button.

machikoyasuda:
@segacy1 
image
Neither of these disabled buttons have sufficient color contrast.

It won't affect the home page, but it will affect the job post page. The job post page has a `disabled` button when the job is no longer open, like this: https://compiler.la/jobs/executive-assistant

image


```
 • Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 4.3:1. Recommendation:  change text colour to #171717.
   ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail
   ├── #apply
   └── <a class="d-inline-block monospace btn btn-primary disabled" id="apply" href="https://forms.clickup.com/8631512/f/87d6r-2087/GUOF0LDD4SN58BC27T"> Apply Now </a>

 • Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 4.3:1. Recommendation:  change text colour to #171717.
   ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail
   ├── html > body > main > div > div > div:nth-child(27) > a:nth-child(1)
   └── <a class="disabled monospace btn btn-primary" target="_blank" href="/capabilities">Capabilities Statement</a>

 • Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 4.3:1. Recommendation:  change background to #171717.
   ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail
   ├── html > body > main > div > div > div:nth-child(27) > a:nth-child(2)
   └── <a class="disabled monospace btn btn-outline-primary" href="mailto:hello@compiler.la">Get in touch</a>

3 Errors
```

styles/base.css
```diff
@@ -1,18 +1,22 @@
...
     --bs-primary-rgb: 136, 180, 64;
     --bs-secondary: #51B1C7; /* Use with .text-secondary */
     --bs-secondary-rgb: 81, 177, 199;
+    --bs-light: #545454; /* Use with .border-light above Footer - /brand/scales/gray/4 */
+    --bs-light-rgb: 84, 84, 84;
+    --bs-info: #BEB3B3; /* Use with .text-info */
+    --bs-info-rgb: 190,190,190;
     --bs-font-sans-serif: "Roboto", system-ui,-apple-system,"Segoe UI","Helvetica Neue","Noto Sans","Liberation Sans",Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji";
     --bs-font-monospace: "Source Code Pro Bold",monospace;
+    --focus-box-shadow: 0 0 0 calc(2.5rem / 16) var(--brand-primary-black), 0 0 0 calc(4rem / 16) var(--brand-primary-green);
```
machikoyasuda, line 19: This code creates this effect as a box-shadow:
image


How to read this code:
- `0 0 0 calc(2.5rem / 16) var(--brand-primary-black)`: Create a 2.5px-thick black line, the same color as the background color.
-  `0 0 0 calc(4rem / 16) var(--brand-primary-green)`: _Below_ that first line, _starting from the same point_, add another 4px of green line. Only 1.5px of green will actually show up. That's calculated like this: 4-2.5 = 1.5.

styles/base.css
```diff
@@ -39,19 +42,19 @@ li {
...
 
 a {
-    color: var(--green);
+    color: var(--brand-primary-green);
     text-decoration: none;
     transition: 250ms;
 }
 
-li>a {
-    text-decoration: none;
```
machikoyasuda, line 48: No longer necessary because _all_ `a` elements have `text-decoration: none` now.

styles/base.css
```diff
@@ -78,43 +81,60 @@ h4,
...
-/*
-Overwrite Bootstrap's heading font-sizes
-https://stackoverflow.com/questions/5410066/what-are-the-default-font-sizes-in-pixels-for-the-html-heading-tags-h1-h2
-*/
 
 h1 {
     font-size: 2rem;
     margin-top: 1.5rem;
     margin-bottom: 1.5rem;
 }
```
machikoyasuda, line 119: Updating h1, h4 will happen in subsequent PR #37. It will be required for completing #66.

styles/base.css
```diff
@@ -78,43 +81,60 @@ h4,
...
+    font-weight: var(--h2-font-weight);
+    line-height: var(--h2-line-height);
+    letter-spacing: 0;
+    margin: 0;
 }
 
-h3 {
-    font-size: 1.17rem;
+h3,
+.posting h2 {
```
machikoyasuda, line 131: Since each Job Post is created from Markdown, and we can't declare CSS classes in Markdown, this `.posting h2` declares that all H2s on a job post should actually use `h3` styles, which is what Figma calls for.

styles/base.css
```diff
@@ -131,34 +151,74 @@ h6 {
...
+    --bs-btn-disabled-border-color: var(--button-primary-disabled);
+    --bs-gradient: none;
+}
+
+.btn-primary:focus,
+.btn-primary:focus-visible,
+.btn-outline-primary:focus,
+.btn-outline-primary:focus-visible {
+    background-color: var(--bs-btn-bg);
+    border-color: var(--bs-btn-border-color);
```
machikoyasuda, line 221: This is a really, really neat nesting CSS Variable feature.

The underlying Figma logic calls for: For `btn-primary` and `btn-outline-primary`, the background color on focus should be the same as the default background color. Same with the border color. Bootstrap code has these values set to the _hover_ colors instead - which is brighter - instead of the default color.

In the design, the background color for Primary is green, and the background color for Outline is black. 

In the code, this is how those values are set:

```css
.btn-primary {
    --bs-btn-bg: var(--button-primary-default);
    --bs-btn-border-color: var(--button-primary-default);
}

.btn-outline-primary {
    --bs-btn-bg: transparent;
    --bs-btn-border-color: var(--button-primary-default);
```

In this code below, I'm able to set the background-color to be the `var(--bs-btn-bg)` _in each class's context_. Even though `var(--bs-btn-bg)` is set to _different_ values for `.btn-primary` and `.btn-outline-primary`, the CSS figures out what value is being scoped. Saves time and lines of code.

```
.btn-primary:focus,
.btn-primary:focus-visible,
.btn-outline-primary:focus,
.btn-outline-primary:focus-visible {
    background-color: var(--bs-btn-bg);
    border-color: var(--bs-btn-border-color);
}
```

machikoyasuda:
Added a bunch of explanatory comments.

@thekaveman @angela-tran Ready for review.

Ideally, both this PR and the #106 PR can go out at the same time, so that the Home Page looks in sync with the new Footer.

jobs.html
```diff
@@ -9,11 +9,11 @@
...
 
-            <h2 class="mt-5 fs-6">Open roles</h2>
-            <ul id="open" class="list-unstyled text-decoration-none"></ul>
+            <h2 class="mt-5 fs-6 pb-1 mb-2">Open roles</h2>
+            <ul id="open" class="list-unstyled d-flex flex-column gap-2 text-decoration-none"></ul>
 
-            <h2 class="mt-5 fs-6">Past roles</h2>
-            <ul id="closed" class="list-unstyled"></ul>
+            <h2 class="mt-5 fs-6 pb-1 mb-2">Past roles</h2>
+            <ul id="closed" class="list-unstyled d-flex flex-column gap-2"></ul>
```
machikoyasuda, line 16: Use `d-flex flex-column gap-2` to create a 8px vertical gap (as in, padding) between the list items. `d-flex flex-column` sets the links to flow vertically, rather than horizontally and tells `gap` to create the gap vertically.

thekaveman:
@machikoyasuda 

> The design calls for removing the Capabilities Statement button. But since we don't have the https://github.com/compilerla/compiler.la/pull/44 yet, we don't have any other way for people to get to it. Should I keep the button there?

Let's leave both buttons there for now, until we have the nav from #44.

/cc @AnthonyRollins

styles/base.css
```diff
@@ -78,43 +82,60 @@ h4,
...
 
-.btn-link {
-    font-size: var(--button-text-font-size);
+/* Headlines */
+
+:root {
+    --h2-font-size: calc(32rem / 16);
+    --h2-font-weight: 500;
+    --h2-line-height: 120%;
+    --h2-font-size: calc(24rem / 16);
```
machikoyasuda, line 103:
```suggestion
    --h3-font-size: calc(24rem / 16);
```

machikoyasuda:
@segacy1 There's a Netlify preview link now, so you can check out how the buttons look:

machikoyasuda:
Funny mobile browser quirk: Mobile Safari thinks these codes are phone numbers

F1CC815F-3136-4DDB-86A9-A21E9608CA7F

machikoyasuda:
Rebased with latest `main`

segacy1:
Hi @machikoyasuda - 2 quick notes from reviewing on Netlify

1. Could we remove the drop shadow from the nav? 
2. Could we vertically center the content on the page between nav and footer? 

I'll make sure that's reflected in Figma as well

machikoyasuda:
@segacy1 These 2 fixes are now out here: https://deploy-preview-107--compilerla.netlify.app/

segacy1 (approved):

assets/compiler-la--pattern-sm.svg
thekaveman, line 1: Can we please avoid using `compiler-la` as an ID or engineering string whenever possible? Either `compiler` or `compiler-llc` if you need a longer version. For this is could even be like `brand-pattern` or `shapes-brand-pattern` or something.

thekaveman (changes requested):

assets/compiler-la--pattern-sm.svg
machikoyasuda: Got it. I put the company name in there for SEO reasons 😅

machikoyasuda: But Compiler LA is not the official company name anymore right? It's Compiler LLC.

machikoyasuda: renamed https://github.com/compilerla/compiler.la/pull/107/commits/52b9c6697edb4214d0b6c283caf39c05b7c718f4

angela-tran (commented):
The focus ring is such a simple thing, yet makes the site feel more polished! I like it.

I noticed the focus ring around our logo is a little weird - is that an easy fix? 
image

machikoyasuda:
@angela-tran https://github.com/compilerla/compiler.la/pull/107/commits/615c4bbc9b49725bc7023025fc86930b59a96862

image

Works just by adding `d-inline-block` to the parent `a` tag.

angela-tran (approved):
Looks good to me 👍

assets/compiler-la--pattern-sm.svg
machikoyasuda: @thekaveman need a re-review for this! thank u

thekaveman (approved):

PR #108, opened 2023-08-21
Fix: Jobs Page, Pill

### LINKED ISSUE #102: Jobs Page Styling Fixes
Hello @machikoyasuda @angela-tran! I looked at the Jobs page and only have a few small notes. 

**Jobs Page**

Links

- My intention for this section was actually to have past roles unlinked and truncated (which is why they are white). I realize that wasn’t actually discussed with the team, so I’ve updated this in Figma to represent a primary active link. They should be primary green, bold,  and only underline on hover. I've created a design to mirror what's live (no open roles)
Screenshot 2023-08-16 at 2 20 15 PM

- Let’s make subscribe in the last line of the page green without an underline to match the other links (updated in figma)

Page Label

- Please change the color of word “careers” in the label to white
- Label should have 4px top/bottom padding, 8px left/right padding

**Job Posting Page**

Per @machikoyasuda [note in Slack](https://compilerla.slack.com/archives/C04BN95SH0S/p1692126596390079?thread_ts=1691772972.709519&cid=C04BN95SH0S), there will be some inconsistencies in the padding and order of past roles. I can take another look if we have a new posting or make a template.

machikoyasuda:
fix #102 

## What this PR does
- Pill: Pill font is now white color, with padding of 4px/8px.
- App: Body background is now darker.
- Links: Links are green, and underline on hover.

image
image

## Notes
**- Border width:** Figma calls for a very thin border, of 0.2px, but both Chrome and Firefox will only allow a border as thin as 0.5px.
**- Font weight:** I think the font weights that Figma calls for are 200 over what it actually needs to be. (Or put it another way, the code expects a font weight of 500 when Figma says 700.) I confirmed this visually. I'm not sure why this is happening. This pill looks more correct with this weight than what is on Figma.

## Accessibility check
The changes in this PR bring the page up to 100% on Axe. Currently, it's failing for color contrast on production.

image

### DIFF
--- a/jobs.html
+++ b/jobs.html
@@ -5,7 +5,7 @@ title: Jobs with Compiler
 <div class="jobs container mt-5">
     <div class="row">
         <div class="col-md-8 offset-md-2 mt-5">
-            <span class="pill rounded p-2">Careers</span>
+            <span class="pill">Careers</span>
             <h1>Compiler is a woman-owned software consultancy that’s passionate about making government tech solutions
                 accessible for all.</h1>
 
--- a/jobs.html
+++ b/jobs.html
@@ -9,10 +9,10 @@ title: Jobs with Compiler
             <h1>Compiler is a woman-owned software consultancy that’s passionate about making government tech solutions
                 accessible for all.</h1>
 
-            <h2 class="mt-5 fs-5">Open roles</h2>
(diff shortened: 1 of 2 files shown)


machikoyasuda:
@angela-tran @thekaveman Ready for review

angela-tran (approved):
LGTM 👍

PR #113, opened 2023-08-23
Fix: Jobs/Posting alignment on Desktop

### LINKED ISSUE #112: Bug: Jobs/Post page alignment off
image

The Job Post page should be the same column width as the Jobs page, but it's not. It's missing a `container` wrapper.

machikoyasuda:
fixes #112 

The design calls for the column-width alignment of the Jobs page and the Posting pages to be the same. They currently are not the same on Prod. This PR makes them the same:

image


angela-tran (approved):
Looks good!

PR #5, opened 2022-06-03
Feat: first-pass index.html

angela-tran:
Closes #2 

### Summary
Adds a basic `index.html` file which mirrors our business cards with Compiler branding. Currently not utilizing any Jekyll features yet.

PR #116, opened 2023-08-23
Fix: Font weights

### LINKED ISSUE #37: Update fonts and colors

### LINKED ISSUE #101: Investigate font-weight discrepancy (and fix if needed)
> This can be changed in future PR's b/c it's not a big deal - but I'm not sure if this is actually rendering font-weight 600 (Semi Bold) font: I typed in `CAREERS` here at weight 600, size 12px here (https://fonts.google.com/specimen/Source+Code+Pro?preview.text=CAREERS&preview.text_type=custom) and compared it with the Netlify. I think we need to add the SemiBold font file in this directory https://github.com/compilerla/compiler.la/tree/main/fonts and make sure the SemiBold file is getting rendered.
> 
> image

_Originally posted by @machikoyasuda in https://github.com/compilerla/compiler.la/issues/87#issuecomment-1668308714_

machikoyasuda:
fixes #101 
closes #37 
depends on #107 to be merged first

## What this PR does
- Use preloaded externally-hosted Google web font instead, with specific font weights
- Changes font weights

## Font weights

This is a little chaotic 😰 but the font-weights in Figma don't always match up with what we want on the browser for CSS.

  | css | figma
-- | -- | --
**a** | **500** | **700**
h1 | - | -
h2 | 700 | 700
h3 | 700 | 700
**h4** | **600** | **800**
**button** | **600** | **700**
**pill** | **400** | **600** 

I just had to visually test different weights and get it approved with Segacy. Generally, the smaller font sizes don't need to be as heavy, font-weight-wise. Whereas, H2 and H3 are fine as 700, H4, A or buttons are too heavy, and have to have decreased font-weights. This makes it even more important that all type use an already-defined type style (a, h1, h2, h3, h4, button, body/p) - so we know it's correct. I haven't added H1 yet simply b/c none of our pages have an H1 yet - that will come with the About page creation.

In case you're wondering why this discrepancy exists:
- https://forum.figma.com/t/why-does-a-font-weight-in-figma-seem-lighter-than-the-same-weight-in-the-browser/2207/2
- In the past, before Figma existed, I used to use something like this (https://forum.figma.com/t/font-in-browser-seem-bolder-than-in-the-figma/24656/6) to make fonts look better on retina screens (back when that was a thing), but that doesn't change the rendering for me, and it has accessibility/browser compatibility issues.

### H2 class
image

image

image

### H3 class
image
image

### H4 class
image

### Pill
image

### Links
image

### Buttons
image
image


machikoyasuda:
@segacy1 How do these font weights look to you?

https://deploy-preview-116--compilerla.netlify.app/
https://deploy-preview-116--compilerla.netlify.app/jobs
https://deploy-preview-116--compilerla.netlify.app/jobs/associate-consultant-scheduling-coordinator

angela-tran:
Hi @machikoyasuda - here is the reasoning for having our own copy of the fonts: https://github.com/compilerla/compiler.la/pull/5#issuecomment-1156851344. The TLDR is at the bottom of the linked article:

> Google Fonts resources will be redownloaded for every website, regardless it being cached on the CDN. Self-host your fonts for better performance. The old performance argument is not valid anymore.

machikoyasuda:
@angela-tran Yea I can move it back. I just did it this way for now so I can quickly get all the weights up and make a Netlify Preview, so Segacy can confirm what weights we need.

The font weight values from Figma aren't 1:1 with what we actually need for the browser.

machikoyasuda:
@angela-tran Google argues otherwise 😆 https://web.dev/font-best-practices/#using-self-hosted-fonts - I want to get the font-weights down first before doing all the font performance stuff.

angela-tran:
@machikoyasuda Oh, interesting find. If there's all this additional nuance to it, then I'm fine with just going with loading from Google Fonts.

segacy1:
Weights look good to me! One q - are the weights on the two buttons on home the same? Hard for me to tell visually because of the different color contrasts

machikoyasuda:
@segacy1 Yea the buttons are the same font-weight. Isn't it trippy that they look so different b/c of the colors?!?!

segacy1:
@machikoyasuda truly! In that case, this looks great to me. Thank you!

segacy1 (approved):

_layouts/default.html
```diff
@@ -26,6 +23,10 @@
     <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-rbsA2VBKQhggwzxH7pPCaAqO46MgnOM80zW1RWuH61DGLwZJEdK2Kadq2F9CUG65" crossorigin="anonymous">
     <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.3/dist/js/bootstrap.bundle.min.js" integrity="sha384-kenU1KFdBIe4zVF0s0G1M5b4hcpxyD9F7jL+jjXkk+Q2h455rYXK/7HAuoJl+0I4" crossorigin="anonymous"></script>
 
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&family=Source+Code+Pro:ital,wght@0,500;0,600;0,800;1,700&display=swap" rel="stylesheet">
```
machikoyasuda, line 28: On preconnect: https://web.dev/font-best-practices/#preconnect-to-critical-third-party-origins

styles/base.css
```diff
@@ -1,25 +1,27 @@
-@import url("./fonts.css");
-
 :root {
```
machikoyasuda, line 1: Preliminary performance comparison between externally-hosted and self-hosted fonts. Keep in mind, the externally-hosted fonts are request more font-weights now - 2 for Roboto and 4 for Source Code Pro (vs. 1 for Roboto and 2 for Source Code Pro). 

This PR vs. Production, ran tests 2 times in a row:

image
image

It's hard to get consistent Performance results (like look at how Bootstrap differs across tests) - but it seems like loading 4 weights of Source Code Pro via externally-hosted is still faster than loading 2 weights of self-hosted.

Technically, we're not using weight 800 for Source Code Pro yet, but I kept it in b/c it'll be used for the H1, which is used in the About pages coming up.

machikoyasuda:
@angela-tran @thekaveman This is now ready for eng review. @segacy1 has approved it design-wise.

thekaveman (commented):
Does this need to be rebased? I feel like some of these commits were from the Homepage PR?

machikoyasuda:
@thekaveman Rebased

_includes/certs.html
```diff
@@ -1,6 +1,6 @@
 {% for cert in site.data.certs %}
 <div class="pe-lg-5 pt-lg-5 mt-lg-1 pb-5">
-    <h4 class="text-secondary pb-md-2">{{ cert.name }}</h4>
+    <span class="h4 text-secondary pb-md-2">{{ cert.name }}</span>
```
thekaveman, line 3: Nice catch on the out-of-order headings.

I still think these should all be real heading elements though, looks like `<h2>` is probably the right level? Since this cert stuff is all important content we want to be high up in search results.

thekaveman (changes requested):

_includes/certs.html
machikoyasuda: It would be `<h2>` level for the home page, but since these certs are in the Footer, which is on every page - the `<h2>` would not be accurate for the job list page or the job index page. So I'm not sure what would be the right heading element, unless we made different footers for every template.

thekaveman: I think it is still accurate? Like you said the footer is on every page, and every page has an `<h1>`, so we can always have one or more `<h2>` in the footer.

machikoyasuda: Figured it out! Fixed here: https://github.com/compilerla/compiler.la/pull/116/commits/bbeaa067915e2238a4bced6d5f237781b808fcae

machikoyasuda:
Rebased main and took care of comments. Ready for re-review @thekaveman @angela-tran

thekaveman (approved):

PR #117, opened 2023-08-24
Nav

### LINKED ISSUE #110: Development work for Nav
We skipped the Nav for the Jobs/Career Milestone, so we are adding it here for the Home Page milestone.

### Menu trigger icon button:

- For focus, let's give this the same treatment as the buttons: 1.5px green border with 4px padding between border and hamburger icon
- For hover, let's have it go to brand/primary/gray. Not really noticeable but I don't think this really needs a hover state? - lmk if you think otherwise!

### Menu close icon button:

- Same as trigger buton

### Menu link:

- Underline on hover, remain white.
- Standard H2 (40/32)

machikoyasuda:
closes #110 

## What this PR does
- Add the nav to desktop and mobile
- Remove Capabilities Statement button from Home page
- Remove `Back to Jobs` link from Job Post page

## Nav
- 3 links: Home, Jobs, Capabilities Statement. Uses standard H2 style, link and underline in white. Cap Statement opens in a new tab.
- Header logo: Add width/height for performance
- Toggle button and X button, created by background SVGs in CSS, have their own custom focus, hover effects
- Accessibility: Aria labels for X, hamburger icon
- All Header and Nav css is in `nav.css`, HTML is in `nav.html`

### On load
image

image

### Hover
image
image

### Focus
image
image

### Hover
image

### Focus
image


machikoyasuda:
@segacy1 Could you send me a tightly-cropped SVG for the nav background image, without the extra text? You might be able to see from here: https://github.com/compilerla/compiler.la/pull/117/files#diff-4e0c8e8d8ccf59da38242b6386c09d7de9f07bf73063b415c98bd53239cd1517 that the 2 files we have both have a lot of extra margin and padding. It would be easier for us code-wise to put the files in if the files are tightly cropped.

styles/nav.css
```diff
@@ -0,0 +1,74 @@
...
+.btn-close {
+    color: var(--white);
+    background: transparent url('data:image/svg+xml,<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg"><g clip-path="url(%23clip0_1358_20250)"><rect x="26.2341" y="28.0208" width="32" height="2" transform="rotate(-135 26.2341 28.0208)" fill="white"/><rect x="3.60669" y="26.6066" width="32" height="2" transform="rotate(-45 3.60669 26.6066)" fill="white"/></g><defs><clipPath id="clip0_1358_20250"><rect width="32" height="32" fill="white"/></clipPath></defs></svg>') center/2rem auto no-repeat;
+    width: 2rem;
+    height: 2rem;
+}
+
+.btn-close:hover {
+    background: transparent url('data:image/svg+xml,<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg"><g clip-path="url(%23clip0_1358_20250)"><rect x="26.2341" y="28.0208" width="32" height="2" transform="rotate(-135 26.2341 28.0208)" fill="%23878787"/><rect x="3.60669" y="26.6066" width="32" height="2" transform="rotate(-45 3.60669 26.6066)" fill="%23878787"/></g><defs><clipPath id="clip0_1358_20250"><rect width="32" height="32" fill="%23878787"/></clipPath></defs></svg>') center/2rem auto no-repeat;
+}
```
machikoyasuda, line 47: I used this site https://www.svgbackgrounds.com/tools/svg-to-css/ to convert the SVG files from Figma into CSS data:images. Then, for the hover color change, I changed `fill="white"` to `fill="%23878787"`.

styles/nav.css
```diff
@@ -0,0 +1,74 @@
...
+
+@media (min-width: 992px) {
+    :root {
+        --modal-header-padding: 0 calc(24rem / 16);
+    }
+}
+
+.modal-header {
+    padding: var(--modal-header-padding);
+    height: calc(72rem / 16); /* Match header height */
```
machikoyasuda, line 21: This keeps the X / menu icons vertically centered.

styles/nav.css
```diff
@@ -0,0 +1,74 @@
...
+    opacity: 20%;
+    margin: calc(-10rem / 16);
+}
+
+@media (min-width: 992px) {
+    .nav-footer-pattern {
+        background-image: url("/assets/illustrations/footer-lg.svg");
+        margin: 0;
+    }
+}
```
machikoyasuda, line 74: Need better image file to avoid the negative margin thing.

styles/nav.css
thekaveman: Would the same `mask-image` approach from #121 work here and be simpler?

machikoyasuda: Just a personal preference - but I actually prefer this approach -- technically this approach is compatible with more browsers. This approach doesn't work for #121 b/c the SVG is already encoded and the code doesn't have an easy `fill` value to change.

thekaveman (approved):
Looks great!

PR #119, opened 2023-08-25
About

### LINKED ISSUE #66: Development work for About page

machikoyasuda:
closes #66 

## What this PR does
- Applies Copy from here: https://docs.google.com/spreadsheets/d/16ehalxbx9DoTqOPAHkVUC_dIYa4PlP4RohVrZSNz0Gg/edit#gid=0 to the designs
- Adds data file `team.yml` with team member info
- Adds 1 testimonial quote, company values, recent work
- CSS: Adds some new color variables, adds H1, text and border utility classes
- CSS: Animate the bus and bike

## Typography
- H1 - Title
- H2 - Subtitles for each section
- H3 - names (names from testimonials, names of team members)
- 🟢 Type hierarchy test passing

## A11y
- 🟢 Aria: Passing for Collapse button
- 🟢 Color contrast: Passing


machikoyasuda:
@segacy1 @AnthonyRollins Here's how the site looks like now on the preview site: https://deploy-preview-119--compilerla.netlify.app/about

- @segacy1 What do you think about the font sizes of the testimonial quotes? 

Without Gillian's quote, it looks like this at the bottom:

image

machikoyasuda:
@AnthonyRollins @segacy1 

This is a question for the both of you.

First, @AnthonyRollins - This section has links to benefits.calitp.org, https://ods.calitp.org/ and https://www.interoperablemobility.org/ - Do you want these linked? Or should they be unlinked? Do you want anything to be linked in the Customer Success section?
image

Second, @segacy1 - The green link on the blue bg is not color contrast sufficient. We need to change the background or the link color.

machikoyasuda:
Status: This PR is now ready for content review from @anthonyrollins, design review from @segacy1 

After content and design review are done, it'll be ready for engineering review.

AnthonyRollins:
@machikoyasuda To answer your questions --

> Without Gillian's quote, it looks like this at the bottom:

@machikoyasuda Let's remove Gillian's quote for now, as we still need her final approval to use it on our website. 

> @AnthonyRollins @segacy1
> 
> This is a question for the both of you.
> 
> First, @AnthonyRollins - This section has links to benefits.calitp.org, https://ods.calitp.org/ and https://www.interoperablemobility.org/ - Do you want these linked? Or should they be unlinked? Do you want anything to be linked in the Customer Success section? image

I want to keep the links in there, but now I notice that the designs don't have links built into those sections. @segacy1 let me know if that works for the design, if not, we can eliminate them. 

@segacy1 for the bus and bike - I was under the impression that we were adding an animation that would move the bus and bike across the screen as the user scrolls down the page, is that not the case?

machikoyasuda:
Updated with a bunch of changes from @segacy1 and @AnthonyRollins:
- Second quote hidden
- First quote now with new colors
- Recent work section with new colors
- Links on white backgrounds are now black with black underline on hover and black square ring on focus
- Bus and bike are animated. Bus has a pause. Bike has a 3 second delay.
- About page link added to Nav


Ready for re-review from @segacy1 and @AnthonyRollins

AnthonyRollins:
**Recent Work Section**
Screenshot 2023-08-29 at 3 21 58 PM

@segacy1 for the branding - with the background now white in the `Recent work section,` should the links now have a green color to match the `jobs page`?

I might change the content of the quote section in the future, but it looks good for now.

**Bus & Bike Animation**
It looks like the animation just moves across the page no matter if we scroll down or not, @segacy1 I believe we wanted to scroll as the user goes down the page, but maybe you talked with Machiko and discussed this would work better. I don't really have a preference here, but I do love that they are moving.

machikoyasuda:
Need to fix: Tablet/mobile widths
025587FA-4D06-4965-9959-D5EC1BCD702C

segacy1:
**Animation**
@machiko two notes for the animation: 
1. Let’s add 4 more seconds to the bus "stop"
2. Let’s not have animation ease out on exit. Can it be linear/leave the screen at the same speed? 

@Anthony yes, originally we wanted the animation to be activated by user scroll. The only issue with this animation for these particular icons is that when you scroll back up, they’d need to move backward on the tracks. If that doesn’t feel too weird to the group, though, I’m open to revisiting!


**Links**
@AnthonyRollins our primary green is accessible on black backgrounds but _not_ on any of our light backgrounds. We’d need to decide between either: 

1. Creating a dark green color for links that can work on light backgrounds
2. [What we did] Changing the link style so that it’s not color-dependent. Now a link is indicated by being underlined and bold by default.

I went with two because even if we do create a dark green color that’s accessible, it still might not look great on all light backgrounds we may use in the future (white, blue, green, etc). I figured this would be easier to manage down the road. 

If the links don’t look particularly “link-y” (lol) I'm think we can up the font weight so the bold is a bit bolder and it's clearer it's a link. What do you think?

segacy1:
@machikoyasuda a few more things:

1. I can't tell if the "ready to work together" section at the bottom has vertically centered text. Could you also decrease that section's top/bottom padding to what is in figma and make the text H3?
2. Could you have the height of the "our philosophy" section match the "recent work" section?
3. For some reason, this title text looks less bold than the rest of the text on the page. Could you confirm weight, please? 

Screenshot 2023-08-30 at 10 40 47 AM

AnthonyRollins:
> 2\. [What we did] Changing the link style so it’s not color-dependent. Now a link is indicated by being underlined and bold by default.

@segacy1 Makes sense to me. I just wanted to check on this. 

> @anthony yes, originally we wanted the animation to be activated by user scroll. The only issue with this animation for these particular icons is that when you scroll back up, they’d need to move backward on the tracks. If that doesn’t feel too weird to the group, though, I’m open to revisiting!

I hear that concerned about it moving backward on the tracks, so let's keep it moving on its own like it is now. We don't need to rethink this animation if you are okay with it moving independently from the user scrolling with your changes about the bus stopping and ease.

machikoyasuda:
Ready for a final design review @segacy1 & code review @angela-tran @thekaveman

segacy1:
This looks great @machiko!

Just a couple more small things:

- Could you make sure links are underlined by default? 
- All buttons should be full-width like on home
- The animation is almost there. Only thing to change is to match the bus speed after the stop to the speed it has before the stop. It's much faster on the way out. 
- Since it's an uneven number here, could we have one more person show up in the team section on mobile (this is more of a nice to have, not a big deal)
IMG_9655

- On mobile, social media links show up as white boxes for me (I'm on chrome):
IMG_9656

machikoyasuda:
Latest updates:
- Links are now underlined, and not underlined on focus and hover. 
- Pill is font-weight 600
- Buttons are mobile are width 100%
- Footer icon bug confirmed is not a bug
- Top H1 section is 16px shorter in total height
- Tracks padding issue is fixed - no more extra padding above tracks
- Confirmed: Can't really change the number of team member illos on mobile from 5 to 6, so it'll stay the same

_data/team.yml
```diff
@@ -0,0 +1,64 @@
...
+
+- name: Laney Mangan
+  title: Associate Consultant, Customer Support Specialist
+  image: "laney-mangan.png"
+  classes: ""
+
+- name: Sarah Henry
+  title: UX Designer
+  image: "sarah-henry.png"
+  classes: "collapseMore collapse"
```
thekaveman, line 29: In a future iteration I'd like to see us move presentation logic out of this file, and do something with e.g. Javascript to shuffle and show 5 images on page load.

This will make it easier for new team members to make an edit to this file to get themselves listed, and would make that page a little more fun/dynamic.

about.html
```diff
@@ -0,0 +1,150 @@
...
+<section class="row d-flex align-items-center h-100vh-minus-nav position-relative">
+    <div class="container">
+        <div class="offset-lg-1 col-lg-7 py-5 my-5 py-md-0 my-md-0">
+            <span class="pill">About Us</span>
+            <h1 class="mt-4 mb-3">We make government services accessible for all</h1>
+            <p class="mb-4 text-white">At Compiler we’re dedicated to building open-source, human-centered, secure, solutions to support the delivery of government services that increase equity of opportunity.</p>
+            <a class="btn btn-primary mb-4 mb-md-0" target="_blank" href="/capabilities">Read our Capabilities Statement</a>
+        </div>
+        <span class="animate-bus">
+            <img alt="Illustration of a bus" src="/assets/illustrations/bus.svg">
```
thekaveman, line 15: I love this little guy! In a future iteration, it would be nice if we could force it fully above the fold, so the entire bus is visible when someone lands on this page:

image

styles/base.css
```diff
@@ -273,3 +320,126 @@ h6 {
...
+    }
+}
+
+.animate-bus {
+    right: -62px;
+    bottom: 0;
+    position: absolute;
+    -webkit-animation: move-left 22s linear infinite;
+    animation: move-left 22s linear infinite;
+}
```
machikoyasuda, line 382: The animation requirements I got from Segacy were: Move for 6 seconds, pause for 4 seconds, move for 4 seconds for the total animation, then have an 8 second pause in between loops. Each total cycle is 22 seconds, which includes the pause. Lastly, she asked that the bus travel at the same speed during both its 2 segments.

This was the math I did to arrive at these percentages. The first numbers (`27%, 45%`) are an expression of _times_, as percentages of the total 22 second loop. The last number (`60%`), is how far the object travels along the x-axis, so 60% means 60% of the viewport width. Because the overall _speed_ of the two "move" rows are the same, at 10 (percent of viewport width per second), the bus should look like it's going the same speed throughout the page.

So the bus takes 6 seconds (from 0 to 27%) to go 60% of the width, then pauses there for 4 seconds (which is expressed as a total 45% elapsed time). Then, it starts again from the 60% width point to the end in 4 seconds. Lastly, it pauses _off_ the page, expressed as `calc(100vw + width of the bus);` for 8 seconds. 

duration |   | time in seconds | time in percent |   | location start | location end | total distance | speed
-- | -- | -- | -- | -- | -- | -- | -- | --
  |   | sum | (time / total) % |   | x-axis % | x-axis % | (start-end) | (distance/seconds)
6 | move | 6 | 27.27272727 |   | 0 | 60 | 60 | 10
4 | pause | 10 | 45.45454545 |   | 60 | 60 | 0 | 0
4 | move | 14 | 63.63636364 |   | 60 | 100 | 40 | 10
8 | pause | 22 | 100 |   | 100 | 100 | 0 | 0

I hope I did the math right 😅

thekaveman (approved):
LGTM code wise!

about.html
machikoyasuda: It should be above the fold fully:

image

machikoyasuda: But we can't have it 100% always above the fold, *IF* the height is not tall enough to fix all the text -- or else, the bus will run into the button. So that's why there's the a pretty tall `min-height` there for mobile screens, to avoid this situation:
image
image

thekaveman: Right, I totally get that. My screenshot is just from a laptop screen, is there a way to shorten that min-height without impacting mobile? It is definitely above the fold on a normal desktop monitor.

styles/base.css
thekaveman: This is intense 🤯

about.html
machikoyasuda: Try this: https://github.com/compilerla/compiler.la/pull/119/commits/8fa16ff205ebc85b17e555e745490ae5232ba013

Try this at your regular browser text size level, and then try boosting up to 200% font size -- it shouldn't cause the bus/button to clash.

thekaveman (approved):

about.html
thekaveman: Looks great!
image

PR #126, opened 2023-09-15
Fix: button font-weight

angela-tran:
The current font-weight for buttons on the live site is 600, but our Figma designs have it as 700.

image

image

This PR changes it to 700.

PR #127, opened 2023-09-15
Feat: 404

### LINKED ISSUE #120: Design for 404 Page
Design a branded design to replace Netflify's default 404:
Screenshot 2023-08-28 at 11 31 15 AM

angela-tran:
Closes #120 

Changes made on top of #126 so need to wait for that to be merged


thekaveman:
@angela-tran it's not a problem for mergeability, but this branch contains the commit that came in from #126 -- it should be rebased on `main` interactively to drop that duplicate commit.

angela-tran:
> @angela-tran it's not a problem for mergeability, but this branch contains the commit that came in from #126 -- it should be rebased on `main` interactively to drop that duplicate commit.

Ah, yep -  thank you

404.html
```diff
@@ -0,0 +1,19 @@
...
+        </div>
+        <div class="offset-lg-3 col-lg-6 mb-md-4">
+            <p class="mb-3 pb-3 text-info">The page you’re looking for doesn’t exist or was removed. We suggest you go
+                back home.</p>
+            <div class="d-flex flex-column flex-lg-row mb-5 pb-5">
+                <a class="btn btn-primary" href="/">Go back home</a>
+            </div>
+        </div>
+    </div>
+</div>
```
thekaveman, line 19: The ordering of elements here is slightly off.

We always want `.row > .col-x` e.g. columns are direct descendants of rows.

Here `.container` is sitting in the middle, it should be on the outside wrapping the `.row`.

https://getbootstrap.com/docs/5.2/layout/grid/

thekaveman (changes requested):

404.html
thekaveman: I'm noticing this by going back and forth to a found / not found page, and seeing the widths are different.

E.g. https://deploy-preview-127--compilerla.netlify.app/jobs (top) vs. https://deploy-preview-127--compilerla.netlify.app/nope (bottom)

image

angela-tran: Yeah, I think it would be better if they had the same offset, but the design calls for them to be different. @segacy1 do you have thoughts on this?

image

image

segacy1: Let's match it to jobs! Will update in figma now

angela-tran: > The ordering of elements here is slightly off.

There's a [`.container-fluid`](https://github.com/compilerla/compiler.la/blob/main/_layouts/default.html#L51) from the default layout that wraps the `.row`.

> We always want `.row > .col-x` e.g. columns are direct descendants of rows.

Yeah, I also noticed that in [lines 6-7](https://github.com/compilerla/compiler.la/pull/127/files#diff-128cc35baaf40bac8ae4d2932ca6e009311e8279f03f258e9cc83966f1d5b32eR6-R7) we don't have a `.row` in between the `.container` and `.col`. I can fix that. I was going off what's in [`index.html`](https://github.com/compilerla/compiler.la/blob/a7abe22129f32bf7311b9d117a24c2d216a4f30d/index.html#L7-L8), which is also missing the `.row`.

segacy1:
It looks like there might a little extra vertical padding between elements. On both desktop and mobile total padding should be:

Between—
icon & header: 24px
header & body: 8px
body & button: 32px

Screenshot 2023-09-15 at 12 42 59 PM

404.html
thekaveman: @segacy1 if possible, every page should have the same offset / width in the content area. It is quite jarring going between pages when that changes. (While also noting that some pages obviously need more space, e.g. for a hero image or something)

404.html
```diff
@@ -0,0 +1,21 @@
+---
+layout: default
+---
+
+<div class="row">
+    <div class="container my-5 py-5">
+        <div class="row">
```
machikoyasuda, line 7: After #135 and #134, I believe these two lines can be removed.

404.html
```diff
@@ -0,0 +1,21 @@
+---
+layout: default
+---
+
+<div class="row">
+    <div class="container my-5 py-5">
```
machikoyasuda, line 6: After #134 and #135 are merged, these two lines can be removed.

404.html
```diff
@@ -0,0 +1,21 @@
...
+        <div class="row">
+            <div class="offset-lg-2 col-lg-7 mt-md-4">
```
machikoyasuda, line 8: After https://github.com/compilerla/compiler.la/pull/134 and https://github.com/compilerla/compiler.la/pull/135 are merged, this line can be changed to `<div class="offset-lg-1 col-lg-7">`

404.html
```diff
@@ -0,0 +1,21 @@
...
+                <img alt="" width="40" height="40" src="assets/alert.svg">
+                <h1 class="h2 mb-3 pt-4 mt-3">Something’s missing!</h1>
+            </div>
+            <div class="offset-lg-2 col-lg-6 mb-md-4">
+                <p class="mb-3 pb-3 text-info">The page you’re looking for doesn’t exist or was removed. We suggest you
+                    go back home.</p>
+                <div class="d-flex flex-column flex-lg-row mb-5 pb-5">
```
machikoyasuda, line 15: After https://github.com/compilerla/compiler.la/pull/134 and https://github.com/compilerla/compiler.la/pull/135 are merged, I believe the `mb-5` and `pb-5` can also be removed.

machikoyasuda:
I took over this PR from @angela-tran, rebased it with the latest `main`, updated some styles and now it is ready for review: https://deploy-preview-127--compilerla.netlify.app/404

@thekaveman

thekaveman (approved):
LGTM

PR #128, opened 2023-09-15
Feat: add artist credit for team illustrations

### LINKED ISSUE #125: Credit artist for the employee illustrations on About page
Please credit @worksofharte (see her Instagram url below) for the illustrations on the About page. Perhaps something like "Illustrations by @worksofharte" underneath the illustrations?
https://www.instagram.com/worksofharte/

angela-tran:
Closes #125

### DIFF
--- a/styles/base.css
+++ b/styles/base.css
@@ -336,6 +336,7 @@ header {
 
 .fs-8 {
     font-size: calc(12rem / 16) !important;
+    letter-spacing: 1.2px !important;
 }
 
 .bg-w-100 { /* Full-width background color, goes beyond gutter */
--- a/styles/base.css
+++ b/styles/base.css
@@ -3,6 +3,7 @@
 :root {
     --white: #FFFFFF;
     --text-black: #1C1C1C;
+    --text-grey: #BEBEBE;
     --brand-primary-green: #88B440;
     --brand-primary-black: #1C1C1C;
     --brand-primary-cyan: #51B1C7;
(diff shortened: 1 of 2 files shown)


styles/base.css
```diff
@@ -278,6 +279,18 @@ header {
     }
 }
 
+/* Credit */
+.credit {
+    color: var(--text-grey);
+    font-family: var(--bs-font-monospace);
+    font-size: 12px;
```
thekaveman, line 286: We use `rem` everywhere else `font-size` is declared, should probably do so here as well?

angela-tran: Thanks for catching 🙏 I grabbed this CSS from Figma and thought I had it all cleaned up - missed this font unit

segacy1:
@angela-tran could you make the weight of "worksofharte" the same as "team illustrations by"? It should be bolder

thekaveman:
> @angela-tran could you make the weight of "worksofharte" the same as "team illustrations by"? It should be bolded

This may be because the font-weight is specified as `800` but we only bring in `500-700`: https://github.com/compilerla/compiler.la/blob/main/_layouts/default.html#L31

angela-tran:
I think it's because the font weight from `a` element styles ends up overriding 😳 

image

I'll fix it so the link is just as bold as the text

machikoyasuda:
Finishing up this PR by adding proper font-weight to the link. Refactored CSS.

Ready for review @angela-tran @thekaveman

thekaveman (approved):

Issue #123, opened 2023-08-30
Clean up file structure, devcontainer


thekaveman:
- [ ] Move all site content / assets into a `src/` directory
- [ ] Update the Jekyll config to build from this directory
- [ ] Update the devcontainer build to account for new directory if needed
- [ ] Update the devcontainer base image to python 3.11
- [ ] Make devcontainer use a non-root user
- [ ] `.prettierrc` to control indentation and other formatting

machikoyasuda:
- [ ] use pyproject.toml (https://github.com/cal-itp/benefits/blob/dev/pyproject.toml)

PR #134, opened 2023-09-26
Horizontal alignment

### LINKED ISSUE #131: Redesign: Page horizontal alignment
- Desktop - Homepage, About, Jobs index, Jobs post, 404: Main content should be col-lg-7 offset-lg-1. Mobile remains the same.

machikoyasuda:
closes #131

### DIFF
--- a/_layouts/job_post.html
+++ b/_layouts/job_post.html
@@ -3,70 +3,68 @@ layout: default
 title_prefix: "Jobs with Compiler:"
 ---
 
-<div class="container posting">
-  <div class="row">
-    <div class="offset-md-2 col-md-7 col-12 mt-5 pt-5 mb-5 pb-5">
-      <h1 class="h2 mt-4 pt-2 mb-5 pb-2">{{ page.title }}{% if page.type %} ({{ page.type }}){% endif %}</h1>
+<div class="row posting">
+    <div class="offset-lg-1 col-lg-7 mt-5 pt-5 mb-5 pb-5">
+        <h1 class="h2 mt-4 pt-2 mb-5 pb-2">{{ page.title }}{% if page.type %} ({{ page.type }}){% endif %}</h1>
 
-      {{ content }} {% if page.apply_link %}
-      <a
+        {{ content }} {% if page.apply_link %}
+        <a
         class="d-inline-block monospace primary-btn"
         id="apply"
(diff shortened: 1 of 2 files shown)


_layouts/job_post.html
```diff
@@ -3,70 +3,68 @@
...
-      apply.setAttribute("aria-disabled", "true");
+    if(open_date > now || close_date < now) {
+        const apply = document.getElementById("apply");
+        if (apply) {
+            apply.classList.add("btn", "disabled");
+            apply.setAttribute("aria-disabled", "true");
+        }
     }
-  }
 </script>
```
machikoyasuda, line 70: Re-indented this page as 4 space.

_layouts/job_post.html
```diff
@@ -3,70 +3,68 @@
 title_prefix: "Jobs with Compiler:"
 ---
 
-<div class="container posting">
```
machikoyasuda, line 6: This parent `container` is no longer necessary b/c the default layout has a `container-fluid` parent already.

_layouts/job_post.html
machikoyasuda: This app doesn't have a `pyproject.toml` file or anything setting the indentation (_I think_).

thekaveman: I saw your comment on the clean up issue #123 

I think we want a `.prettierrc` file that specifies the spacing, since Prettier is our default formatter for HTML, JS, CSS -- something like:

```json
{
  "tabWidth": 2,
  "useTabs": false,
  "printWidth": 120
}
```

machikoyasuda: Can do this is a separate PR. The indentation on some of these HTML files are extremely messed up.

thekaveman: > Re-indented this page as 4 space.

Hmm I thought we used 2 though?

machikoyasuda: I thought we used 2 too, but NONE of the pages at least..when I look at my VS Code use 2 spaces. Getting this on all pages:
image

But that confused me b/c apparently Prettier's default is 2: https://prettier.io/docs/en/options.html#tab-width

So I'm not really sure where the 4 is coming from.

thekaveman (approved):
LGTM but I think there will be a merge conflict between this and #136

PR #135, opened 2023-09-27
Vertical alignment

### LINKED ISSUE #130: Redesign: Page vertical realignment
## Alignment
- Desktop - Homepage, About, Jobs index, Jobs post, 404: Distance between bottom of header and top of brandmark should be 144px. 
- Mobile - Homepage, About, Jobs index, Jobs post, 404:  Distance between bottom of header and top of brandmark should be 104px.

machikoyasuda:
requires #134 
closes #130 

## What this PR does
- Adds CSS that declares the following: The first `row` in the `main` element with the `container-fluid` class (`main.container-fluid > .row:first-child`) has a margin-top/bottom of 144px or 104px depending on the device/window width.
- The `default.html` layout has a `<main class="container-fluid">`. So this means that whenever a new page is created, the developer _MUST_ start the page with a `<section class="row">`, in order to get the correct vertical alignment.
- Why not add the top/bottom margins to `<main class="container-fluid">` directly? Because that's not what Figma calls for, and it would create design errors on the About page.

## All new pages that use the Default layout should have

```html
<section class="row">
  <div class="col-lg-7 offset-lg-1">
      # this is element that will get the 144/104px top/bottom margin
  </div>
</section>
```

### DIFF
--- a/_layouts/job_post.html
+++ b/_layouts/job_post.html
@@ -4,8 +4,8 @@ title_prefix: "Jobs with Compiler:"
 ---
 
 <div class="row posting">
-    <div class="offset-lg-1 col-lg-7 mt-5 pt-5 mb-5 pb-5">
-        <h1 class="h2 mt-4 pt-2 mb-5 pb-2">{{ page.title }}{% if page.type %} ({{ page.type }}){% endif %}</h1>
+    <div class="offset-lg-1 col-lg-7">
+        <h1 class="h2 mb-5 pb-2">{{ page.title }}{% if page.type %} ({{ page.type }}){% endif %}</h1>
 
         {{ content }} {% if page.apply_link %}
         <a
--- a/_layouts/job_post.html
+++ b/_layouts/job_post.html
@@ -3,47 +3,46 @@ layout: default
 title_prefix: "Jobs with Compiler:"
 ---
 
-<div class="row posting">
+<div class="row posting baseline-padding">
     <div class="offset-lg-1 col-lg-7">
         <h1 class="h2 mb-5 pb-2">{{ page.title }}{% if page.type %} ({{ page.type }}){% endif %}</h1>
 
-        {{ content }} {% if page.apply_link %}
-        <a
-        class="d-inline-block monospace primary-btn"
-        id="apply"
-        href="{{ page.apply_link }}"
-        >
-        Apply Now
-    </a>
-    {% endif %}
+        {{ content }}
 
-    <p>
-        <em>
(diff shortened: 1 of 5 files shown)


jobs.html
```diff
@@ -3,23 +3,22 @@
...
 
-            <p class="my-5">
-                Missed a past opportunity? <a href="http://eepurl.com/h6qTKL">Subscribe</a> to our mailing list, as we
-                are always looking for great candidates.
-            </p>
-        </div>
+        <p class="mt-5 mb-0">
+            Missed a past opportunity? <a href="http://eepurl.com/h6qTKL">Subscribe</a> to our mailing list, as we
+            are always looking for great candidates.
+        </p>
```
machikoyasuda, line 21: Correcting confusing indentation

_layouts/job_post.html
```diff
@@ -4,46 +4,45 @@
...
+                rel="noopener noreferrer">
+                    #banthebox
+                </a>
+            </em>
+        </p>
+
+        <div>
+            <a class="monospace lg-link" href="/jobs">Back to Jobs</a>
+        </div>
+    </div>
```
machikoyasuda, line 45: Correcting confusing indentation

thekaveman:
@machikoyasuda 

> All new pages that use the Default layout should have...

Would it make sense to build this into the default layout? So we don't have to repeat this layout code in other pages?

machikoyasuda:
@thekaveman No, because some off-pages like the Blog wouldn't work with it.

machikoyasuda:
@thekaveman I wrote the docs there to just remind people that the page should start with <div class="row"> and then columns within it. It can be easy to forget there's a container already and start the page with <div class="container"> but that would mess up how the column children get rendered.

thekaveman:
@machikoyasuda there is a conflict here, please rebase / resolve when you get a chance.

machikoyasuda:
@thekaveman Rebased + resolved conflicts 🙏

machikoyasuda:
@thekaveman This is ready for review 👋

thekaveman (commented):
I think the animation on the About page needs to be adjusted? This is what I'm seeing on the preview site -- the bus is below the bike and the box rather than on top of the box:

image

machikoyasuda:
@thekaveman thanks for the catch! fix is here: https://deploy-preview-135--compilerla.netlify.app/about

thekaveman (approved):
This looks so much better thank you!! :pray: :+1:

PR #171, opened 2024-02-15
fix: fix jekyll build bug

### LINKED ISSUE #170: Bug: Cannot build app locally
Running build step leads to:

```
An error occurred while installing sass-embedded (1.70.0), and Bundler cannot
continue.

In Gemfile:
  jekyll was resolved to 4.3.3, which depends on
    jekyll-sass-converter was resolved to 3.0.0, which depends on
      sass-embedded
```

machikoyasuda:
fixes #170 

Same bug as https://github.com/cal-itp/mobility-marketplace/issues/523

from https://github.com/jekyll/jekyll-sass-converter/issues/161

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source "https://rubygems.org"
 
 gem "jekyll", "~> 4.3.3"
+gem "jekyll-sass-converter", "2.2.0"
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Confirmed this fixes the issue for me! :pray:

PR #204, opened 2024-06-26
Chore: add issues to project board workflow

### LINKED ISSUE #8: Automatically add issues to the Project Board
This will be helpful for the team as we start to build out the backlog - so everything is captured on the board!

See https://github.com/cal-itp/benefits/blob/dev/.github/workflows/add-to-project-issues.yml for an example of a workflow that does this.

Related:

* compilerla/compiler-admin#20
* compilerla/data-donuts#124

lalver1:
Closes #8 

I think the only missing part for this PR is creating the `GH_PROJECTS_TOKEN` secret. I haven't been able to find a way to do this, maybe I'm missing permissions @thekaveman ? Or if you can create it from your account that works too.

### DIFF
--- a/.github/workflows/add-to-project-issues.yml
+++ b/.github/workflows/add-to-project-issues.yml
@@ -1,0 +1,14 @@
+name: "Project triage: Issues"
+
+on:
+  issues:
+    types: [opened, transferred]
+
+jobs:
+  add-to-project-issues:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@main
+        with:
+          project-url: https://github.com/orgs/compilerla/projects/${{ vars.GH_PROJECT }}
+          github-token: ${{ secrets.GH_PROJECTS_TOKEN }}


.github/workflows/add-to-project-issues.yml
```diff
@@ -0,0 +1,14 @@
...
+    types: [opened, transferred]
+
+jobs:
+  add-to-project-issues:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@main
+        with:
+          project-url: https://github.com/orgs/compilerla/projects/${{ secrets.GH_PROJECT }}
+          github-token: ${{ secrets.GH_PROJECTS_TOKEN }}
```
angela-tran, line 14: @lalver1 I think to create this, you'd log in to GitHub as @compiler-bot and generate a new PAT. Then you can add that as a GitHub secret. 

I think an organization-level secret makes sense? cc @thekaveman

lalver1: I see, thanks @angela-tran!

thekaveman: Thanks @angela-tran!

Org secret definitely makes sense for the PAT generated from the @compiler-bot for `GH_PROJECTS_TOKEN`. That's how we do it in Cal-ITP. That org secret should be restricted and only available to certain repos (those where we will auto-add issues to the project).

I think for the `GH_PROJECT` (i.e. the ID of the project) we can make that an org variable, no need for it to be secret.

angela-tran (approved):

thekaveman (changes requested):
@lalver1 it looks like these are repository variables/secrets, and not org variables/secrets.

Can you create org-level data with the bot account? Maybe it doesn't have permissions.

angela-tran:
@thekaveman We've been chatting about the permissions in Slack: https://compilerla.slack.com/archives/C04BN95SH0S/p1719505167402449?thread_ts=1719442336.376809&cid=C04BN95SH0S - let us know if you think we should go a different direction from what we discussed there

thekaveman (approved):

lalver1:
Thanks @angela-tran , I was about to comment but the Slack link is great 😁

Issue #175, opened 2024-03-04
Add Blog feed


rdela:
https://github.com/jekyll/jekyll-feed 

> ## Installation
> 
> Add this line to your site's Gemfile:
> 
> ```ruby
> gem 'jekyll-feed'
> ```
> 
> And then add this line to your site's `_config.yml`:
> 
> ```yml
> plugins:
>   - jekyll-feed
> ```
>

rdela:
These articles both cover non-plugin approaches: 

- [RSS and Atom feeds | CloudCannon Academy](https://learn.cloudcannon.com/jekyll/rss-and-atom-feeds/)
- [RSS Feeds and Jekyll | John Vincent](https://www.johnvincent.io/jekyll/rss-feed-with-jekyll/)

thekaveman:
Thanks for this suggestion @rdela! We'll look at getting this implemented with `jekyll-feed` soon.

thekaveman:
Blog feed is now published here: [`https://compiler.la/blog/feed.xml`](https://compiler.la/blog/feed.xml)

rdela:
A feed is better than no feed. Yet left wondering, why set [`excerpt_only: true` ](https://github.com/compilerla/compiler.la/pull/202/files#diff-ecec67b0e1d7e17a83587c6d27b6baaaa133f42482b07bd3685c77f34b62d883R13
) - [jekyll-feed readme](https://github.com/jekyll/jekyll-feed?tab=readme-ov-file#excerpt-only-flag)

To me the idea of a feed reader is to be able to read articles from multiple sources in one place, so I would prefer to have article body (“post content”) in the feed. 

@thekaveman @angela-tran

thekaveman:
Thank you for your feedback @rdela! We've updated the feed to publish the full post body.

rdela:
Splendid! https://mastodon.social/@rdela/112689358545751339

PR #207, opened 2024-06-26
Refactor: send the full post body in blog feed

thekaveman:
Per https://github.com/compilerla/compiler.la/issues/175#issuecomment-2192359853


lalver1 (approved):
Looks good!
It's interesting, using this [Firefox feed reader add-on](https://addons.mozilla.org/en-US/firefox/addon/feedbroreader/) I'm able to see the post content both when `excerpt_only` is `true` or `false`.

PR #216, opened 2024-06-27
Feat: randomize team on About on page load

### LINKED ISSUE #199: Randomize display order of staff on About
Every time the page (re)loads, the order of the staff under `Meet the team` should be randomized from the default order of the items in the [`yaml` data file](https://github.com/compilerla/compiler.la/blob/main/src/_data/team.yml)

thekaveman:
Closes #199

### DIFF
--- a/src/about.html
+++ b/src/about.html
@@ -67,7 +67,7 @@ description: We build open-source, human-centered, secure, agile solutions to su
                     <span class="text-white pill border-0 ps-0 pb-md-5 mb-md-3 pb-4 mb-3 d-inline-block">Our philosophy</span>
                 </div>
                 <div class="offset-lg-1 col-lg-5">
-                    <h2 class="text-white mb-3 mb-md-4">Empower government workers </h1>
+                    <h2 class="text-white mb-3 mb-md-4">Empower government workers</h2>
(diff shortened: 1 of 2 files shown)


src/about.html
```diff
@@ -150,3 +153,27 @@ <h3 class="text-dark">Ready to work together?</h3>
...
+        // and shuffle
+        while (currentIndex != 0) {
+            let shuffleIndex = Math.floor(Math.random() * currentIndex);
+            currentIndex--;
+            [members[currentIndex], members[shuffleIndex]] = [members[shuffleIndex], members[currentIndex]];
+        }
+        // then override the flex order with new shuffled index
+        // hiding all but the first 5
+        members.forEach((m, i) => {
+            m.style.order = i;
```
angela-tran, line 171: I did not know about flex [`order`](https://developer.mozilla.org/en-US/docs/Web/CSS/order) - that's cool!

src/about.html
```diff
@@ -150,3 +153,27 @@ <h3 class="text-dark">Ready to work together?</h3>
...
+            if (i > 4) {
+                m.classList.add("collapse", "collapseMore");
```
angela-tran, line 173: This also means if Javascript is disabled, all the pictures will show ❤️

angela-tran (approved):
This is awesome, nice!!

PR #218, opened 2024-06-28
Feat: implement top section of /our-work page

### LINKED ISSUE #206: Implement top "Our Work" section
This styles on this section are similar to the top section on `/about`, so you can reference it when implementing this one.

https://github.com/compilerla/compiler.la/blob/2c251ef784f78069d9c8efc702762360f9ff3898/src/about.html#L6-L18

Screenshot for what's in scope here:
image

Check against Desktop and Mobile mockups in Figma.

lalver1:
Closes #206

I'm not sure if we should keep the bus animation
```
    <span class="animate-bus">
      <img alt="Illustration of a bus" src="/assets/illustrations/bus.svg" />
    </span>
```
in this section, I removed it but can put it back in if needed.

### DIFF
--- a/src/our-work.html
+++ b/src/our-work.html
@@ -4,7 +4,18 @@ title: Our Work
 ---
 
 <section class="row position-relative">
-  <!-- Our Work -->
+  <div class="container">
+    <div class="offset-lg-1 col-lg-7 py-5 my-5 py-md-0 my-md-0">
+      <span class="pill">Our Work</span>
+      <h1 class="mt-4 mb-3">Building solutions that empower government services</h1>
+      <p class="mb-4 text-white">
+        We are committed to building open-source, accessible solutions to elevate government workers through teaching and training
+        to ensure their success.
+      </p>
+      <a class="btn btn-primary me-lg-3 mb-3 mb-lg-0" href="mailto:hello@compiler.la">Contact us</a>
+      <a class="btn btn-outline-primary" target="_blank" href="/capabilities">Capabilities statement</a>
+    </div>
+  </div>
 </section>
 
 <section class="bg-w-100 position-relative bg-white">


src/our-work.html
```diff
@@ -4,7 +4,18 @@
...
-  <!-- Our Work -->
+  <div class="container">
+    <div class="offset-lg-1 col-lg-7 py-5 my-5 py-md-0 my-md-0">
+      <span class="pill">Our Work</span>
+      <h1 class="mt-4 mb-3">Building solutions that empower government services</h1>
+      <p class="mb-4 text-white">
+        We are committed to building open-source, accessible solutions to elevate government workers through teaching and training
+        to ensure their success.
+      </p>
+      <a class="btn btn-primary me-md-3 mb-3 mb-md-0" href="mailto:hello@compiler.la">Contact us</a>
```
angela-tran, line 15: Two thoughts here:
- I'm not seeing what effect `me-md-3` has on the button, but maybe I'm not testing it correctly. Could you explain why it's needed here?
- I see that `mb-md-0` is similar to [the one on `/about`](https://github.com/compilerla/compiler.la/blob/main/src/about.html#L12). I suggest we move it to the last button (`Capabilities statement`) because right now, the buttons end up stacked on top of each other with no space in between.
   - Side note: to be honest, I'm not 100% sure why `mb-md-0` is being used on the `/about` page. The [commit for it](https://github.com/compilerla/compiler.la/commit/4a3b725a830f81e546f25fa6b5c5571cc1519990#diff-2212b84a2a6ffe239872c76d266680d86f2438391034b2eb19604f17104f6968R12) says it was to handle a window that is too short. In any case, I think it's fine to have here, just on the bottom button instead.

Buttons have no space between them:
image

angela-tran (changes requested):
This is looking great!! Just some small style things to work through

angela-tran:
> in this section, I removed it but can put it back in if needed.

No bus animation needed 👍

src/our-work.html
lalver1: Thanks! Those are very good points! Yep, in general I'm using [breakpoints](https://getbootstrap.com/docs/5.2/layout/breakpoints/) and [margin spacing](https://getbootstrap.com/docs/5.2/utilities/spacing/) to add space between the buttons.

- `me-md-3` adds a 16px (end/right) margin to the 'Contact us' button on a `md` and bigger window. On a smaller than `md` window, the margin is not applied so when the buttons stack up, they are still aligned. But really, `md` is not required since the buttons look aligned when the page renders on a narrow window, I added `md` because without it, the 16px margin is still there, and it looked weird when inspecting the element:
image
But really it should be `me-lg-3` to ensure that we don't see the behavior shown in the picture for medium sized windows too.
- That's true, I thought that `mb-md-0` was the right breakpoint because I only tested a large window and a mobile screen using the 'responsive design mode' browser feature, but I see that there is no space in between when the buttons are stacked on a medium sized window.

So locally I now implemented `me-lg-3 mb-3 mb-lg-0` on the 'Contact us' button to achieve the right spacing (if I'm not mistaken) and I think it means: If on `lg` or bigger, add a 16px right margin; for all screen widths, set a bottom margin of 16px; but if on `lg` or bigger, set no bottom margin. Would this be ok? I can also modify the 'Capabilities statement' button instead but I think we'd need an `mt` class since we'd need to add top margin space instead of bottom margin space.

angela-tran: > But really it should be me-lg-3 to ensure that we don't see the behavior shown in the picture for medium sized windows too.

I see, that makes sense. Thank you!

And for the bottom margin, that all sounds fine

angela-tran (approved):

angela-tran:
I think GitHub's current [degraded performance with Actions / Pull Requests](https://www.githubstatus.com/incidents/9vwllhs2w1kj) is causing the [pre-commit failure](https://results.pre-commit.ci/run/github/494608691/1719594784.ihUFljUbQmCdW_r3169XqA). I'm fine with waiting for it to pass before merging or just merging it.

lalver1:
Ah I was wondering about that. Yep, it's probably related to the degraded performance; I'll wait for it to pass since I don't think this blocks anything.

Issue #212, opened 2024-06-27
Implement "What We've Done" section


angela-tran:
This issue will probably be best done through multiple small pull requests rather than one big pull request.

- Implement cards
   - [x] tags on cards
   - [x] non-image card
   - [x] image card
- [x] Let's have a [data file](https://jekyllrb.com/docs/datafiles/) for the card contents
- [x] "See more projects" button and "See fewer projects" button -> reference team member section from `/about` for how this is done

https://github.com/compilerla/compiler.la/blob/eeeb1ff70c93decf07c5c47a9411edf040d78eeb/src/about.html#L134 

Check against Desktop and Mobile mockups in Figma.

### Screenshots of what's in scope:

<details><summary>Expand</summary>

| Desktop | Mobile |
| --- | --- |
| ![image](https://github.com/compilerla/compiler.la/assets/25497886/40b48519-3d7f-4502-9edf-4398134ffae3) | ![image](https://github.com/compilerla/compiler.la/assets/25497886/0b997927-8825-4fde-8309-3d80471e9e11) |

When expanded:

![image](https://github.com/compilerla/compiler.la/assets/25497886/aaeda528-c170-42f3-9d18-e7d3115c0f21)

</details>

_moved screenshots rounding decision to PRs (e.g. https://github.com/compilerla/compiler.la/pull/220#issue-2380982428)_

PR #220, opened 2024-06-28
Feat: "What We've Done" section - first two cards

angela-tran:
Part of #212 

This PR implements the "What We've Done" section with just the first two cards.

### Dev analysis of spacing and decisions to round:

| Desktop non-image card | Mobile non-image card |
| --- | --- |
|  |  |

### DIFF
--- a/src/our-work.html
+++ b/src/our-work.html
@@ -92,7 +92,7 @@ title: Our Work
         <div class="col">
           <div class="card h-100 text-dark">
             <div class="card-body p-4 m-2 p-lg-5 m-lg-3">
-              <h3 class="card-title fw-boldest">Customer Success Strategy</h3>
+              <h3 class="card-title fw-boldest pb-4 mb-0 pb-lg-2 mb-lg-5">Customer Success Strategy</h3>
               <p class="card-text">
                 Implemented a new Customer Success program, including setting up a Customer Relationship Management (CRM) platform
                 and assigning Account Managers to work with California Transit Agencies.
@@ -110,7 +110,7 @@ title: Our Work
         <div class="col">
           <div class="card h-100 text-dark">
             <div class="card-body p-4 m-2 p-lg-5 m-lg-3">
-              <h3 class="card-title fw-boldest">Operational Data Standard (ODS)</h3>
+              <h3 class="card-title fw-boldest pb-4 mb-0 pb-lg-2 mb-lg-5">Operational Data Standard (ODS)</h3>
               <p class="card-text">
                 Convened the Operational Data Standards (ODS) working group to draft ODS v1, an open standard for describing
                 scheduled transit operations by building on GTFS to include personnel and non-revenue service data.
--- a/src/our-work.html
+++ b/src/our-work.html
@@ -90,25 +90,29 @@ title: Our Work
     </div>
 
     <div class="offset-lg-1 col-lg-10 mb-md-4">
-      <div class="row row-cols-md-1 row-cols-lg-2 g-4">
+      <div class="row row-cols-1 row-cols-lg-2 g-4">
         {% for featured_work in site.data.featured_work limit: 2 %}
         <div class="col">
           <div class="card h-100 text-dark">
-            <div class="card-body p-4 m-2 pb-5 mb-3 px-lg-5 mx-lg-3 pt-lg-5 mt-lg-5 pb-lg-4 mb-lg-6">
-              {% if featured_work.tags %}
-              <div class="row row-cols-1 row-cols-lg-auto g-2 g-lg-1">
-                {% for tag in featured_work.tags %}
-                <div class="col">
-                  <span class="pill pill-tag" aria-hidden="true">{{ tag }}</span>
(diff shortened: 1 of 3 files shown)


angela-tran:
Rebased onto latest `feat/our-work`

angela-tran:
Sorry, found one more spacing thing to fix

lalver1 (approved):
Looks good!

angela-tran:
🙏

thekaveman (commented):
This is going to be tricky, but I think ideally these `Outcome` section headers / sections should line up (independent of the size of the description above).

image

thekaveman:
I wonder if @machikoyasuda has any ideas?

src/our-work.html
```diff
@@ -79,7 +79,43 @@ <h2 class="accordion-header" id="our-services-{{ service.id }}">
 </section>
 
 <section class="bg-w-100">
-  <!-- What We've Done -->
+  <div class="container my-5 py-md-5 py-4">
+    <div class="offset-lg-1 col-lg-7">
+      <span class="pill border-0 ps-0 pb-md-5 mb-md-3 pb-4 mb-3 d-inline-block">What we've done</span>
```
machikoyasuda, line 84:
```suggestion
      <span class="pill border-0 ps-0 pb-md-5 mb-md-3 pb-4 mb-3 d-inline-block">What we’ve done</span>
```

machikoyasuda:
> > This is going to be tricky, but I think ideally these `Outcome` section headers / sections should line up (independent of the size of the description above).
> 
> I noticed that too just yesterday while looking at Figma. I agree it'll be tricky... I haven't thought of a way to do it yet. 🤔
> 
> _edit_: maybe we can set a minimum height on things? idk

Yeah, a min-height with a percentage would work - only when the screen width is wide enough / containers are set to be side by side:

image


image

Sidenote: I think there's something funky going on with the responsiveness. At a certain width, the layout goes down to 1 container per row, but if you go further - it goes to 2 containers per row again. 

image

angela-tran:
@machikoyasuda is helping with the alignment of card contents

machikoyasuda:
image
image

src/styles/base.css
```diff
@@ -454,10 +460,15 @@ header {
...
 
+
 @media screen and (min-width: 992px) {
     .w-md-50 {
         width: 50% !important;
     }
+
+    .mb-lg-6 {
+        margin-bottom: calc(96rem / 16) !important;
+    }
```
machikoyasuda, line 471: Could be removed now I think @angela-tran

src/our-work.html
```diff
@@ -95,7 +95,7 @@ <h2 class="mb-4">Featured work</h2>
         <div class="col">
           <div class="card h-100 text-dark">
             <div class="card-body px-lg-5 mx-lg-3 py-lg-5 my-lg-5 pt-3 mt-3 pb-5 pb-3 px-4 px-2">
-              <div class="not-outcome" style="min-height: 70%">
+              <div class="min-h-70">
```
machikoyasuda, line 98: I based this naming off of this class https://getbootstrap.com/docs/5.3/utilities/sizing/#relative-to-the-viewport

src/our-work.html
```diff
@@ -109,7 +109,7 @@ <h3 class="card-title fw-boldest pb-4 mb-0 pb-lg-2 mb-lg-5 pt-2 pt-lg-0 mt-4">{{
                 <p class="card-text pb-2 mb-5">{{ featured_work.description }}</p>
               </div>
 
-              <div class="outcome">
+              <div>
```
machikoyasuda, line 112: Removed this class because it was purely informational for me to see what's in the Outcome section and above it. Had no styles.

src/our-work.html
thekaveman: The docs there say:

> Width and height utilities are generated from the utility API in _utilities.scss. Includes support for 25%, 50%, 75%, 100%, and auto by default. Modify those values as you need to generate different utilities here.

Do we need CSS for the 70% height here?

machikoyasuda: Yea there aren't any min-height percentage classes, only min-height viewport ones. The quote from the Doc refers to height and width classes, not min-height ones.

thekaveman: Hmm ok... but it still looks like we are missing a style rule then? I don't see any CSS for `.min-h-70` and I don't think that class is taking affect:

image

machikoyasuda: Oops - just fixed this with https://github.com/compilerla/compiler.la/pull/220/commits/57d202cafe0c365ae95fb3496e753986103f6b6a

Can you retest?

thekaveman: Looks good :+1:

thekaveman (approved):
Nice team effort here y'all :sunglasses:

src/styles/base.css
angela-tran: Sounds good, I'll do it in one of the next PRs and link it here

PR #221, opened 2024-06-28
Feat: implement "Ready to work together?" section of /our-work page

### LINKED ISSUE #214: Add "Ready to work together?" section
Similar to the one on the `/about` page

https://github.com/compilerla/compiler.la/blob/eeeb1ff70c93decf07c5c47a9411edf040d78eeb/src/about.html#L144-L151

Screenshot of what's in scope:
image

lalver1:
Closes #214

### DIFF
--- a/src/_includes/ready-to-work.html
+++ b/src/_includes/ready-to-work.html
@@ -1,0 +1,10 @@
+<div class="container my-5 py-5">
+  <div class="offset-lg-2 col-lg-8 text-center">
+    {% assign h3-text = "Ready to work together?" %} {% if include.h3-class %}
+    <h3 class="{{include.h3-class}}">{{h3-text}}</h3>
+    {% else %}
+    <h3>{{h3-text}}</h3>
+    {% endif %}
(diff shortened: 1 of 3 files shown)


thekaveman (commented):
Would it be possible to use an include here? Then we can have a single block used for `/about` and `/our-work`?

lalver1:
Yep, that's a good point 👍 
I used an include with a parameter to set the color of the text (because the background of the section in `/our-work` is black but it is white in `/about`) and a conditional tag to set the class.

I could remove the conditional by using the `default` filter:

`<h3 class="{{ include.h3-class | default: "" }}">Ready to work together?</h3>`

but having `<h3 class="">Ready to work together?</h3>` in the rendered HTML of `/our-work` didn't look right.

The include would be much simpler if the background of the section was the same for both `/about` and `/our-work` (no parameters to pass or any conditional tags) but I'm not sure if we should deviate from what we have in Figma?

thekaveman:
I think this is fine. It can be more flexible to avoid sending the specific class as part of the configuration, maybe instead a `light` or `dark` mode where the include then decides which classes to use etc.

But this will work and we can always revisit when we need to use this component again.

thekaveman (approved):

PR #229, opened 2024-07-09
Feat: "What We've Done" - "see more projects" button

angela-tran:
Closes #212

PR #230, opened 2024-07-09
Fix: missing rows cause extra spacing

### LINKED ISSUE #225: Do a final pass on spacing/alignment
It's possible that while implementing the individual sections, we don't get the alignment _between sections_ exactly right.

This ticket should be done last and involves checking the page's spacing / alignment between sections against Figma and fixing any incorrect alignment / spacing.

angela-tran:
Closes #225 

The problem was just some missing `<div class="row">`s. They were causing some sections to not take up the full width that they should. This matches how the About page lays out its sections.

| Before | After |
| --- | --- |
| image |  image |


angela-tran:
Should be reviewed after #229

machikoyasuda (approved):

Issue #200, opened 2024-06-25
Show the navigation bar in Desktop view


thekaveman:
Don't hide behind a hamburger menu.

Needs design.

PR #260, opened 2025-03-31
Proof-of-concept: desktop navbar

angela-tran:
This is just a rough proof-of-concept for #200. We don't have to stick with anything in this PR; it simply serves as a way to spark conversation on:
- what design / styling we want
- code implementation.

The approach here is to set `d-lg-none` on the existing modal so that it is not used on large screens. And then a different `div` is used to simply show the navbar links for large screens.

I referenced the second example in Bootstrap's ["Navbar > Responsive behaviors > Toggler" docs](https://getbootstrap.com/docs/5.3/components/navbar/#toggler).

The example had some code to style the link corresponding to the current page, so I incorporated that too.

### Preview

See the preview site for a more complete preview. Here are some screenshots too:

Home page with desktop navbar

Our Work page with desktop navbar


thekaveman:
@angela-tran I'm thinking since this was opened in the Jekyll days, and we haven't made a priority of addressing or coming back to it, we should probably close for now? Perhaps we can redo this work under 11ty when we have some time.

angela-tran:
That sounds good to me @thekaveman, thanks!

PR #267, opened 2025-08-08
Fix: Ruby version

angela-tran:
Similar issue as https://github.com/cal-itp/calitp.org/pull/477 - Netlify gets a 404 when trying to download Ruby 3.1

This PR updates the Ruby version and applies the same standardization for how the Ruby version is set.


thekaveman (approved):
I was able to bring up the devcontainer and run the site locally 👍

PR #289, opened 2026-04-06
chore: install node, 11ty and configure autoformatting

jgravois:
this PR goes a smidge further than https://github.com/cal-itp/calitp.org/pull/589

* updates the python container image to use 3.12 (for consistency)
makes Node.js 24x available in the DevContainer (via mise)
* lays down 11ty
* configures better liquid-aware VS Code autoformatting (analagous to https://github.com/cal-itp/calitp.org/pull/596)

once you've rebuilt the DevContainer:

```bash
$ node -v
v24.14.1

$ npm list
compiler.la@2.0.0 /home/compiler/site
├── @11ty/eleventy@3.1.2
├── @shopify/prettier-plugin-liquid@1.10.2
└── prettier@3.8.1

$ npm start
Wrote 0 files in 0.12 seconds (v3.1.5)
[11ty] Eleventy Fatal Error (CLI):
```

i've configured `11ty` as an alternate production branch for when we get further along with the migration, but this piece is safe to land as-is.

### disclaimer

both in this branch and on `main`, when I make any change to an .html file, prettier forces 2 space indention instead of 4. @angela-tran and i will coordinate on how best to tackle a bulk autoformat as we proceed with next steps for the 11ty migration.


.devcontainer/Dockerfile
```diff
@@ -9,6 +9,16 @@ RUN useradd --create-home --shell /bin/bash $USER && \
 
 WORKDIR /home/$USER/site
 
+# Install mise (to manage Node.js) https://mise.jdx.dev/mise-cookbook/docker.html#docker-image-with-mise
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+ENV MISE_DATA_DIR="/home/$USER/mise"
+ENV MISE_CONFIG_DIR="/home/$USER/mise"
+ENV MISE_CACHE_DIR="/home/$USER/mise/cache"
```
jgravois, line 16: i needed to prepend `/home/$USER` to the mise install directory here because of the way the container sandboxes permissions.

after speaking with @angela-tran with this we agreed to investigate simplifying how this is handled to match calitp.org and mobimart as time allows.

angela-tran (approved):
Looks good! Thanks for getting us started

PR #300, opened 2026-04-09
chore: configure prettier pre-commit step and autoformat entire repo

### LINKED ISSUE #296: Set up pre-commit check and re-format all files
It's unclear if we've ever had a consistent formatting experience in this repo. It's been an issue for a while that if you check out `main` and have VS Code run `prettier` on the templates, there will be formatting changes.

This task is to get us in a good state and help make the 11ty conversion work go more smoothly.

jgravois:
closes #296

* first commit configures prettier in a way that is compatible with pre-commit and converts a few liquid templates to be more explicit.
* second commit lets prettier cook
* third commit turns on pre-commit to prove all is well.
* last commit adds a .git-blame-ignore-revs file

personally i'm fine with prettier and the shopify plugin's defaults, but @angela-tran, feel free to play around and test the effect of making changes to the options for quotes (or anything else) if you'd like.


.vscode/settings.json
```diff
@@ -6,5 +6,8 @@
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
-  "editor.tabSize": 2
+  "editor.tabSize": 2,
+  "[css][html][javascript][json][liquid][markdown][yaml]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  }
```
Scotchester: Doesn't the non-language-specific `"editor.defaultFormatter": "esbenp.prettier-vscode"` line above already have this effect?

src/_includes/job-json.liquid
```diff
@@ -0,0 +1,19 @@
+{% comment %}
+{% endcomment %}
```
Scotchester: Can this be removed?

.pre-commit-config.yaml
```diff
@@ -31,4 +31,7 @@ repos:
   #     - id: prettier
   #       types_or: [css, html, liquid, markdown, yaml]
   #       # it would be ideal to run this on 'pre-commit' too. we just havent figured out how to do it yet
-  #       stages: [commit-msg]
+  #       additional_dependencies:
+  #         - prettier@3.8.1
+  #         - '@shopify/prettier-plugin-liquid@1.10.2'
+  #       types_or: [css, html, javascript, json, liquid, markdown, yaml]
```
Scotchester: Will you execute this TODO as the last commit in this PR?

Scotchester (commented):
A few comments to get us started.

.vscode/settings.json
jgravois: the explicit declaration here is redundant until/except in scenarios where the VS Code user has an alternative formatter configured for individual filetypes (like you did for JSON a couple months back).

.pre-commit-config.yaml
jgravois: that's correct. d639b03cc58dae5573375798aaa2d274cb6ed422 already does.

src/_includes/job-json.liquid
jgravois: indeed it should! in the .html file the code comment had content that asked prettier to avoid autoformatting incorrectly.

now its entirely superfluous. 😅

.vscode/settings.json
Scotchester: Ahhh, right :)

.pre-commit-config.yaml
Scotchester: Ah, I was only looking at the first commit. May I pedantically request that you move this change into a later commit so that it's not ignored with the addition of `.git-blame-ignore-revs`?

jgravois: i wouldn't call that pedantry. what i did initially was probably overly clever.

Scotchester: Persnicketiness, perhaps? 😁

jgravois: hah. i was going to say i'd like to wait until we're reasonably certain we don't want to change any other prettier/shopify defaults before i rewrite history, but there's really no drawback to opening up a follow-up PR and ignoring another rev or two if we decide later that we'd like to keep tweaking.

jgravois: > move this change into a later commit so that it's not ignored with the addition of `.git-blame-ignore-revs`

done in a0031a6

Scotchester (commented):
> i'd like to wait until we're reasonably certain we don't want to change any other prettier/shopify defaults before i rewrite history

I'll preface this by saying, if this is the same config you just implemented on calitp.org, I am totally fine to roll with it, but if we're opening the discussion… skimming through the formatting commit's diff, there are a few things that stand out to me:

- This kind of breaking of tags across different lines remains intensely annoying, but IIRC we couldn't find a reasonable way to avoid it.
  ```html
  <span class="secondary-sans-serif-xs d-inline-block pb-4 mb-3"
    >An inside look at Compiler’s internal projects, interests and our client projects.</span
  >
  ```
- Indendation inside `{% comment %}` blocks. I guess NBD if the comment toggle shortcut in VS Code handles the indentation/deindentation automatically.
- I prefer having the first attribute of an HTML tag on the same line as the tag name and the subsequent attributes lined up with it on the lines below.
- I don't like that it's adding a `/` before the closing `>` of self-closing tags.
- I like the (old school?) convention of not indenting an HTML doc's `head` and `body` tags, but this is the least offensive thing on this list to me.

jgravois:
> if we're opening the discussion...

if anything in that tick list can be mitigated/improved by altering prettier or shopify's defaults, I'm not opposed.

with regard to the tag wrapping, I too recollect that we tried and couldn't find a suitable alternative, but it's not outside the realm of possibility that there's something we missed.

I have to admit that my personal enthusiasm for optimization is low. I'm happy to just accept prettier's defaults, but if anyone else wants to play around with this branch before the PR is merged or create another branch of their own later, all the pieces are present to automate reformatting the entire repo and ensure that the diff is ignored afterward.

Scotchester:
Seems like a decent way to spend the last couple hours of the ride into Chicago :)

As I brought up the docs to dig into what options were available, I noticed this in the VS Code extension's README:

> **Note:** VS Code does not support combined language syntax for `editor.defaultFormatter`. You must set the formatter for each language separately:
> 
> ```json
> // ❌ This will NOT work
> {
>   "[javascript][typescript]": {
>     "editor.defaultFormatter": "esbenp.prettier-vscode"
>   }
> }
> 
> // ✅ Use separate blocks instead
> {
>   "[javascript]": {
>     "editor.defaultFormatter": "esbenp.prettier-vscode"
>   },
>   "[typescript]": {
>     "editor.defaultFormatter": "esbenp.prettier-vscode"
>   }
> }
> ```

I'll see if I can test this out.

Scotchester:
> As I brought up the docs to dig into what options were available, I noticed this in the VS Code extension's README:
> 
> > **Note:** VS Code does not support combined language syntax for `editor.defaultFormatter`. You must set the formatter for each language separately:
> > // ❌ This will NOT work
> > {
> >   "[javascript][typescript]": {
> >     "editor.defaultFormatter": "esbenp.prettier-vscode"
> >   }
> > }
> > 
> > // ✅ Use separate blocks instead
> > {
> >   "[javascript]": {
> >     "editor.defaultFormatter": "esbenp.prettier-vscode"
> >   },
> >   "[typescript]": {
> >     "editor.defaultFormatter": "esbenp.prettier-vscode"
> >   }
> > }
> 
> I'll see if I can test this out.

I haven't completely confirmed this yet, but I DID run across two other things I want to mention quickly:

- `[jsonc]` must be explicitly included in order to have the default formatter apply to JSON with Comments files, e.g., `devcontainer.json` and `.vscode/settings.json`.
- This complication from the [VS Code settings docs](https://code.visualstudio.com/docs/configure/settings#_note-about-multiple-language-specific-settings):
  > If you are using [multiple language-specific settings](https://code.visualstudio.com/docs/configure/settings#_multiple-language-specific-editor-settings), be aware that language-specific settings are merged and precedence is set based on the full language string (for example `"[typescript][javascript]"`) and not the individual language IDs (`typescript` and `javascript`). This means that for example, a `"[typescript][javascript]"` workspace setting will not override a `"[javascript]"` user setting.

Maybe this is referring to the same root issue as the Prettier extension README, but it seems like whatever the case, we'll want to list each language separately.

jgravois:
> it seems like whatever the case, we'll want to list each language separately.

in the wild so far we've only seen a collision for json (you) and html (@angela-tran). personally i'm fine with limiting our overrides to what we've actually seen, but if you prefer to be exhaustive, i don't have a problem with that either.

indexing (commented):
@jgravois I've reviewed this PR, as well as the issue and parent issue, but I don't have anything to add to this convo. Let me know if there is anything in particular on which to focus.

jgravois:
> [indexing](https://github.com/indexing) reviewed on behalf of https://github.com/orgs/compilerla/teams/compiler-comms [35 minutes ago](https://github.com/compilerla/compiler.la/pull/300#pullrequestreview-4198543840)

all good @indexing!

@compilerla/compiler-comms is assigned by default in this repo anytime a change is made in the _posts directory with the assumption being that we're usually either publishing a new blog or making a copy change.

https://github.com/compilerla/compiler.la/blob/dafad11ba46e5f9bb19d97bf1ccef7b24504a060/.github/CODEOWNERS#L3

if you ever see a PR like this that isn't a clear attempt to update the content on the https://compiler.la website, i'd say its safe to ignore. you're also _always_ welcome to ask for clarification before you spend time on a fullblown code review.

Scotchester:
I just did some more testing with [vscode-settings-inspector](https://github.com/ekomateas/vscode-settings-inspector) to confirm that listing them individually works and listing them together does not.

> in the wild so far we've only seen a collision for json (you) and html (@angela-tran). personally i'm fine with limiting our overrides to what we've actually seen, but if you prefer to be exhaustive, i don't have a problem with that either.

It's just such a hidden footgun, and hard to realize exactly what's happening when it goes off, I would prefer to be exhaustive, yes. I could push that change to this branch, if you're cool with that.

jgravois:
>  I could push that change to this branch, if you're cool with that.

of course. you have my blessing! 📿

angela-tran:
> I just did some more testing with [vscode-settings-inspector](https://github.com/ekomateas/vscode-settings-inspector) to confirm that listing them individually works and listing them together does not.

I did some testing and can confirm too that we need to list them individually. Thanks for catching @Scotchester!

angela-tran (approved):
This looks good to me 👍 

I'm good with sticking with prettier's defaults. We can always revisit it, and I think getting through the 11ty conversion work is the higher priority.

Scotchester:
Having satiated my curiosity about the process of updating Prettier in the future, I concur with Angela, let's merge this and move forward.

PR #301, opened 2026-04-30
Feat: blog post - Content Localization

angela-tran:
ClickUp task: https://app.clickup.com/t/86b96889r

PR #302, opened 2026-05-01
chore: convert static site builder from Jekyll to Eleventy

### LINKED ISSUE #287: Convert static site builder from Jekyll to Eleventy
- [x] Install 11ty in the devcontainer - done in #289

jgravois:
closes #287

## Steps to test
Open the two sites side-by-side and compare
* https://11ty--compilerla.netlify.app/
* https://compiler.la/

## `TODO:`
- [x] https://github.com/compilerla/compiler.la/pull/326

## Known discrepancies

1. 11ty adds trailing slashes to routes. ie: https://compiler.la/jobs becomes https://deploy-preview-302--compilerla.netlify.app/jobs/ but requests without the trailing slash should redirect seamlessly
1. each page now includes a `<meta name="generator" content="Eleventy v3.1.6">` tag.
(not sure why we don't currently have a jekyll flavored variant, i was under the impression that it was injected automatically.)
1. `<meta name="og:title">` and `<meta property="og:description">` tags are now injected more consistently
1. bootstrap has been upgraded to 5.3.8
1. the production site includes a `<script src="https://static.cloudflareinsights.com/beacon.min.js...">` tag.
(this is injected by Cloudflare itself and will persist when the 11ty site is deployed.)
1. in this branch `<link type="application/atom+xml" href="/blog/feed.xml">` is an root-relative path instead of a fully qualified domain name. i don't think that's problematic, but if anyone else disagrees, it'd be trivial to make it match.


jgravois:
merging #318 introduced a merge conflict here and #319 should really include the blog post we published in #301 anyway.

as i see it we have two options, either of which feel reasonable:

1. rebase the `11ty` branch on top of `main` 
(and update the SHA listed in the `.git-blame-ignore-revs` file)
2. merge `main` into the `11ty` branch
(and tolerate the merge commit)

i know @angela-tran has expressed (understandable) squeamishness about managing history in long-running branches. if there's a tack that y'all prefer, i will make it so.

angela-tran:
@jgravois - I was thinking about all this (the recent blog post merge into `main` and the 11ty branch) earlier today too so I'm glad you asked. :heart:

My thought on the immediate question is that Option 1 aligns the most with our convention to [rebase rather than merge the base branch](https://github.com/cal-itp/calitp.org/pull/603#issuecomment-4215527702). So yep, option 1 was how I was thinking we should handle it.

My thought on my past comments about managing history on long-lived branches is that those were more about how, if we're working on a feature branch with many interdependent changes coming in from different PRs, I would prefer that all those PRs target the feature branch, rather than some of them targeting `main`. This scenario happened when we were working on the MobiMart Rider Benefits page; from looking at the Rider Benefits design, we found some site-wide adjustments that needed to happen, so there was ambiguity over whether those should target `main` or the feature branch -- that is, one person might think of the changes as tangential clean-up that should be deployed to the live site without waiting on the feature being completed, while another opinion might be that those changes came about because of the feature and therefore are a part of it and should be QA'd along with all of it. I take that latter position because it means we can avoid rebases that feel unnecessary and possibly disruptive if many developers are involved.

I'll link to those comments in case it's helpful context for anyone: 
- https://github.com/cal-itp/mobility-marketplace/pull/894#pullrequestreview-4158175939
- https://github.com/cal-itp/mobility-marketplace/pull/905#pullrequestreview-4183661076

In this case with compiler.la / the blog post, it was much more clear-cut that the blog post was totally independent of our 11ty refactor and needed to be deployed to the live site as soon as it was done. So in this case we have a rebase that cannot be avoided, and that's fine.

That's how I view it at least. I'm happy to discuss this more if anyone disagrees with what I'm suggesting :pray:

jgravois:
thanks for the thorough write-up and including those links. 

> another opinion might be that those changes came about because of the feature and therefore ... should be QA'd along with all of it. I take that latter position

i'm all for avoiding rebases when possible. on that note, i'll wait to rebase this branch on top of `main` until after #319 is reviewed/merged if it's okay with y'all. 🤗

angela-tran:
> i'm all for avoiding rebases when possible. on that note, i'll wait to rebase this branch on top of `main` until after #319 is reviewed/merged if it's okay with y'all. 🤗

That sounds good to me. Hopefully the merge conflicts aren't too bad when you get to the rebase :crossed_fingers: :sweat_smile: :hugs:

Scotchester:
Rebases… can't live with 'em, can't live without 'em; am I right?

(I agree with the proposed course of action.)

jgravois:
> Hopefully the merge conflicts aren't too bad when you get to the rebase 🤞 😅 🤗

luckily they weren't bad 😉

package.json
```diff
@@ -3,16 +3,19 @@
   "version": "2.0.0",
   "description": "Software built by humans, for humans, in LA",
   "author": "Compiler",
-  "license": "AGPL-3.0-or-later",
   "type": "module",
+  "license": "AGPL-3.0-or-later",
```
thekaveman: Random but noticed since I'm using a lot of this work for D+D:

The license here is AGPL, but the [repo LICENSE file](https://github.com/compilerla/compiler.la/blob/main/LICENSE) is Apache 2.0. 

Can we please adjust the license here in `package.json` to match?

thekaveman (dismissed):

package.json
jgravois: good catch. done in da217dc (in #326)

thekaveman: Thanks. I thought about leaving the review there, it was kinda random and didn't align with your other changes (also/mostly the file wasn't visible to comment on) but makes sense to do it there!

jgravois: all good. once that one's merged there's a 50/50 chance we'll need one more 'odds and ends' PR to get the refactor over the line. 😇

angela-tran:
Thanks @jgravois for writing up the known differences between current Jekyll-based compiler.la and our new Eleventy-based compiler.la.

On those notes, I have a quick clarification question (I think it was just a typo) and some details to document for future reference / awareness about the meta description change.

---
> 6. in this branch `<link type="application/atom+xml" href="/blog/feed.xml">` is an absolute url instead of a fully qualified domain name. i don't think that's problematic, but if anyone else disagrees, it'd be trivial to make it match.

Do you mean relative URL, rather than absolute URL?

---

> 3. `<meta name="og:title">` and `<meta property="og:description">` tags are now injected more consistently

To be more specific on this, https://github.com/compilerla/compiler.la/pull/326/changes/d7b68cef7efaabe4b552dd2a5cfb4177a2e744ce is where the change was made to have each page use the default description specified in `src/src.11tydata.json` unless it specifies its own `description` in its front-matter. This means that `/blog`, `/jobs`, `/our-work`, and the 404 page, which each used to not have a meta description, will now show one (the default one). 

<details><summary>
<em>Screenshot showing link previews for live compiler.la `/our-work` and `/jobs` pages vs. the ones from the Netlify preview for this branch.</em></summary>
image

</details>

It could be argued that it's better to have more specific descriptions for each page and not make it so easy to use a generic default description, but given that our page descriptions for all this time have been either the generic description or nothing, "don't let perfect be the enemy of good" feels apt here. So that's to say, I am good with this change.

jgravois:
> Do you mean relative URL, rather than absolute URL?

i'd call `<link type="application/atom+xml" href="./blog/feed.xml">` a relative url, but that's just a naming quibble. the crux of the issue is that it is less explicit in this branch that compiler.la is the canonical host for the feed.

i don't think that's problematic. for example https://11ty.dev's own feed tag looks just like ours, but i still felt like it was worth calling it out.

```html
<link rel="alternate" href="/blog/feed.xml" title="Eleventy Blog" type="application/atom+xml">
```

> It could be argued that it's better to have more specific descriptions for each page

i totally agree with this. i'm happy to treat it as a followup action to create an issue so that we don't forget to craft additional page descriptions after the migration is complete. 🤗

angela-tran (approved):
I did a side-by-side comparison / click-through of every page on compiler.la vs. the preview of this branch and didn't see any regressions.

This looks ready to go to me 👍

Scotchester:
I've always thought of "absolute" URLs _including_ a FQDN, and the term I'm used to for link paths beginning with a `/` is "root-relative".

Scotchester (approved):
Everything looks identical to me! I checked all pages (except that I only checked one blog post).

lalver1:
I visited the new site on my phone too and saw no issues, looks good! 👍

lalver1 (approved):

indexing:
I reviewed all major views in the site, visited a few blog posts, and opened the capabilities statement from both URLs. I didn't note any differences. Looks good to go!

indexing (approved):
Looks good to go!

jgravois:
thank you all. 3 down, 1 to go! 😈🚀😅

PR #313, opened 2026-05-06
chore: use _redirects file instead of a custom template and wire up 404 page

### LINKED ISSUE #308: rewire 404 page and use a _redirects file to serve /capabilities and /retreat

jgravois:
resolves #308

https://deploy-preview-313--compilerla.netlify.app/capabilities == https://compiler.la/capabilities
https://deploy-preview-313--compilerla.netlify.app/retreat == https://compiler.la/retreat
https://deploy-preview-313--compilerla.netlify.app/non-existent == https://compiler.la/non-existent

i did this one by hand 😇


src/404.html
```diff
@@ -1,5 +1,6 @@
 ---
 layout: default
+permalink: /404.html
```
angela-tran, line 3: I feel like I keep learning, forgetting, and then re-learning why we have to do this, so just thought I'd write it out 😅 :

11ty will process `src/404.html` and output `_site/404/index.html`, according to the [default behavior](https://www.11ty.dev/docs/permalinks/#not-index-file-names). To instead have it output at `_site/404.html` so that Netlify uses it as the [custom 404 page](https://docs.netlify.com/manage/routing/redirects/redirect-options/#custom-404-page-handling), we need to specify this [`permalink`](https://www.11ty.dev/docs/permalinks/#changing-the-output-location) key.

With [Jekyll](https://jekyllrb.com/tutorials/custom-404-page/#on-github-pages), the `src/404.html` would be output to the root, so no `permalink` was needed.

angela-tran (approved):
Just a note that the [`_redirects` file](https://docs.netlify.com/manage/routing/redirects/overview/) is specific to Netlify, so with this refactor, we no longer will see redirect behavior when running the site locally.

I think this is fine, we didn't have any issue with this decision when adding redirecting in other projects, e.g. [calitp.org](https://github.com/cal-itp/calitp.org/pull/492/changes). Just wanted to note the functional difference for awareness

src/404.html
jgravois: that is _exactly_ right. 🎯

jgravois:
> with this refactor, we no longer will see redirect behavior when running the site locally.

this is correct too. i think the juice is worth the squeeze, but happy to talk about it further if anyone else has reservations.

PR #314, opened 2026-05-09
chore: refactor jobs collection for 11ty

### LINKED ISSUE #306: rewire /jobs and individual /job/:id pages

jgravois:
closes https://github.com/compilerla/compiler.la/issues/306. copilot planning agent prompt:

> this site is halfway through a conversion from jekyll to 11ty. next i need to refactor the src/_jobs/ collection.
> 1. the _jobs directory no longer needs to be prefaced with an underscore. 
> 2. i think job-json.liquid can be deleted in favor of object literals representing invidual job front-matter generated by 11ty itself.
> 3. the custom logic related to luxon in both the job_post layout template and jobs.liquid files may be unnecessary now too.

unlike my other recent PRs, it required quite a bit of back and forth to debug copilot's work, double check my initial assumptions and get things into mergeable form. it felt a little like shepherding an incredibly speedy junior dev.

at this point i'm happy enough with the result, but i'm also ready to put a pin in my experiments to make sure anyone else that's interested in 11ty is able to get their feet wet.

## steps to test

https://deploy-preview-314--compilerla.netlify.app/jobs/ == https://compiler.la/jobs
https://deploy-preview-314--compilerla.netlify.app/jobs/ux-ui-designer/ == https://compiler.la/jobs/ux-ui-designer/

i double-checked the validation of 'live' job descriptions (locally) by specifying a future close date of individual markdown pages.

```md
---
title: Senior UX/UI Designer
open_date: 2025-07-01
close_date: 2027-07-14
---
```


src/_layouts/job_post.liquid
```diff
@@ -55,19 +55,18 @@ title_prefix: "Jobs with Compiler:"
...
-  luxon.Settings.defaultZone = "America/Los_Angeles";
+  luxon.Settings.defaultZone = 'America/Los_Angeles';
   const DateTime = luxon.DateTime;
   const now = DateTime.now();
 
-  const job = {% include "job-json" job:page %};
-  const open_date = DateTime.fromISO(job.open_date).startOf("day");
-  const close_date = DateTime.fromISO(job.close_date).endOf("day");
+  const open_date = DateTime.fromISO('{{ open_date | date: "%Y-%m-%d" }}').startOf('day');
+  const close_date = DateTime.fromISO('{{ close_date | date: "%Y-%m-%d" }}').endOf('day');
```
jgravois, line 63: 11ty has its own front-matter parser that serves up real JS dates. 

these 🤸 cajole liquidjs into reverting the value _back_ to an ISO string.

angela-tran: Would it work to simply use the JS date as-is? (rather than convert it into an ISO string to then turn back into a `DateTime`)

src/jobs.liquid
```diff
@@ -29,7 +29,7 @@ title: Jobs with Compiler
   const DateTime = luxon.DateTime;
   const now = DateTime.now();
 
-  const all_jobs = {% include "job-json" %};
+  const all_jobs = {{ jobs | jsonify }}
```
angela-tran: I really need to just run through this myself and see 😅 (and I will!), but I'll go ahead and leave this question: what is the purpose of `jsonify` here?

I'm assuming `jobs` is coming from the computed data in `src/jobs.11tydata.js` and that the computed data returns as an array of objects, so I guess I'm not seeing why we'd need to convert it into JSON (if that's even what `jsonify` is doing) since it already is JSON? ~I wasn't able to find docs on `jsonify` so am a little stuck. 🙏~

_edit_: I ended up finding docs, e.g. https://liquidjs.com/filters/jsonify.html, but it wasn't so much that I needed to understand what the filter was doing, but rather what `jobs` actually looks like.

src/jobs.liquid
angela-tran, line 40: Since these aren't coming from frontmatter values, but rather the JSON we construct ourselves in `src/jobs.11tydata.js`, we don't need to do the 🤸 from https://github.com/compilerla/compiler.la/pull/314/changes#r3212536064, is that right?

angela-tran (commented):
This is looking good from my read-through of the code! I had just a few questions.

I'll spend some time running this locally too.

src/jobs.liquid
angela-tran: Ah ok, when I take out `| jsonify`, I see the objects are just spit out one after the other (?) rather than being in an array

```
const all_jobs = [object Object][object Object][object Object][object Object][object Object][object Object][object Object][object Object][object Object][object Object]
```

image

src/_layouts/job_post.liquid
angela-tran: Ah, I see... it's a `luxon.DateTime` that we want, not a JS date. All good, ignore me :smile:

angela-tran (approved):
This looks great @jgravois 💯

src/jobs.liquid
jgravois: > I wasn't able to find docs on `jsonify`

the article below does a _fantastic_ job of documenting the liquidjs `jsonify` filter. 😆 

https://liquidjs.com/filters/jsonify.html

j/k, this would be slightly less opaque if i used the `json` filter instead. (TIL that `jsonify` is apparently just an alias for `json`)

i will say though, despite what the documentation says, neither filter stringifies the input. instead, as you noted, they both make the 11ty computed data available within the script tag as an array of json object literals.

src/_layouts/job_post.liquid
jgravois: it's been awhile, so i'll have to play around with it a little bit more, but i remember that all the casting back and forth between date types was a bit of a rough-edge with respect to the liquidjs templating engine in particular.

src/jobs.liquid
jgravois: that's right. 11ty casts ISO date strings in front-matter as date objects, which is what necessitated the extra cartwheels when passing _back_ to liquidjs. 🤸

src/_layouts/job_post.liquid
angela-tran: Yeah, looks like it ultimately gets rendered as a string version of the JS date if we did
```
const open_date = {{ open_date }};
```

so definitely not what we want. :grin: 

image

jgravois: exactly, i _wish_ something like this would work:
```js
const open_date = DateTime.fromJSDate({{ open_date }}).startOf('day');
```

an alternative that does work is 👇 , but that doesn't feel like an improvement.
```js
const jsDate = new Date('{{ open_date }}');
const open_date = DateTime.fromISO(jsDate).startOf('day');
```

src/jobs.liquid
jgravois: i rebased this commit on top of the `11ty` branch and took the opportunity to swap in the canonical filter and add a code comment.

43c3188

src/_layouts/job_post.liquid
angela-tran: Interesting that that works even though the docs for `fromISO` say it takes a String and that it would also work with the `fromObject` and `fromJSDate` methods. :woman_shrugging:  I agree with keeping it as-is

src/jobs.liquid
angela-tran: Good call on the swap and the comment! Looks great

src/_layouts/job_post.liquid
jgravois: yeah, its weird. `{{ open_date }}` is neither a [`String`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) nor a genuine [`Date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)

PR #319, opened 2026-06-05
chore: rewire blog and RSS feed for 11ty

### LINKED ISSUE #293: Configure 11ty to generate RSS feed
In #202, we added an Atom feed using the `jekyll-feed` plugin. The 11ty-built site should maintain this behavior. 

I haven't verified for sure that this 11ty plugin will give us what we want, but it's a place to start - https://www.11ty.dev/docs/plugins/rss/

### LINKED ISSUE #307: rewire /blog and individual blogs
for this one we'll need to create a mechanism for including blog drafts during local development but not in production builds.

https://github.com/compilerla/compiler.la/blob/dafad11ba46e5f9bb19d97bf1ccef7b24504a060/.vscode/tasks.json#L26-L45

luckily, the 11ty doc has a handy snippet we can use for that:
https://www.11ty.dev/docs/config-preprocessors/#example-drafts

jgravois:
closes #293
closes #307

old: https://compiler.la/blog/2026/remix-case-study
new: https://deploy-preview-319--compilerla.netlify.app/blog/2026/remix-case-study/

old: https://compiler.la/blog 
new: https://deploy-preview-319--compilerla.netlify.app/blog/ <sup>1</sup>

✅ [Valid Atom 1.0 feed](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fdeploy-preview-319--compilerla.netlify.app%2Fblog%2Ffeed.xml)

<sup>1</sup> the team picture won't match and [/blog/2026/content-localization](https://deploy-preview-319--compilerla.netlify.app/blog/2026/content-localization) will be missing until/unless we get in sync with `main`.


src/blog/pretty-atom-feed.xsl
jgravois, line 1: this file is an optional (but snazzy) addition that makes the raw RSS feed human readable. ✨ 

https://deploy-preview-319--compilerla.netlify.app/blog/feed.xml

Image

prior art: [11ty/eleventy-base-blog/.../eleventy.config.js#L66](https://github.com/11ty/eleventy-base-blog/blob/50a9cb8e818e857fcd30ce7734fe060ef2600551/eleventy.config.js#L66)

Scotchester (approved):
LGTM!

src/blog/blog.11tydata.js
```diff
@@ -0,0 +1,8 @@
+// /blog/:year/:slug/
+const permalink = ({ page: { date, fileSlug } }) => `/blog/${date.getFullYear()}/${fileSlug}/`;
+
+export default {
+  layout: "blog_post",
+  permalink,
+  tags: ["posts"],
```
thekaveman, line 7: Couple non-blocking questions about `tags` here:

1. Is this currently used anywhere? I wasn't seeing references to it. Why do we need it?
2. If we later wanted to apply tags to certain posts, e.g. `data-donuts` related to my comment yesterday about syndicating posts across sites, or to build some kind of search/filter experience, could we still do that on a per-post basis? Would they merge with this global list, or how does that work?

src/blog/pretty-atom-feed.xsl
```diff
@@ -0,0 +1,88 @@
...
+                  </linearGradient>
+                </defs>
+                <rect width="256" height="256" rx="55" ry="55" x="0"  y="0"  fill="#CC5D15"/>
+                <rect width="246" height="246" rx="50" ry="50" x="5"  y="5"  fill="#F49C52"/>
+                <rect width="236" height="236" rx="47" ry="47" x="10" y="10" fill="url(#RSSg)"/>
+                <circle cx="68" cy="189" r="24" fill="#FFF"/>
+                <path d="M160 213h-34a82 82 0 0 0 -82 -82v-34a116 116 0 0 1 116 116z" fill="#FFF"/>
+                <path d="M184 213A140 140 0 0 0 44 73 V 38a175 175 0 0 1 175 175z" fill="#FFF"/>
+              </svg>
+              Web Feed Preview
```
thekaveman: Why isn't `Compiler's Blog` the `<h1>` title element here? 

I note that is the value of the `<title>` element, and I reckon they should align?

src/blog/blog.11tydata.js
jgravois: this is a great question because its a bit opaque. `tags` in 11ty are the mechanism by which content is organized into collections.

https://www.11ty.dev/docs/collections/

i noticed that right now blog posts are assigned one or more categories:

https://github.com/compilerla/compiler.la/blob/e717571f5b8263019926430d6f68d73a189add7a/src/_posts/2024-01-24-devcontainer-platform-agnostic-team.md#L9-L12

unless i'm missing something, we aren't doing anything in the UI with that information yet, but i'm confident that it'd be straightforward to introduce filtering/search and only syndicate a subset of the blog posts to D+D when we're ready to tackle that.

edit:
tbh, i'm not sure whether it would make more sense to consolidate the `tags` and the `categories` to do it or leave them seperate. i'd guess either option would be workable.

src/blog/pretty-atom-feed.xsl
```diff
@@ -0,0 +1,88 @@
...
+            </h1>
+            <h2><xsl:value-of select="atom:feed/atom:title"/></h2>
+            <p><xsl:value-of select="atom:feed/atom:description"/></p>
+            <p>This preview only shows titles, but the actual feed contains the full content.</p>
```
thekaveman: #175 was the original issue (submitted by an outside reader) that prompted us to publish a feed.

In the same issue, the requestor made the point that publishing the full post body in the feed was more desirable, as it lets folks read the post in their preferred reader app/interface. I don't see a strong reason to change this behavior.

thekaveman (dismissed):

src/blog/blog.11tydata.js
Scotchester: > this is a great question because its a bit opaque. `tags` in 11ty are the mechanism by which content is organized into collections.
> 
> https://www.11ty.dev/docs/collections/

And to connect one more dot for you @thekaveman, since I had the same question when reviewing: the `posts` collection is what's used for the main blog listing:

https://github.com/compilerla/compiler.la/blob/76859b5396d6f5a86f5c1234fc0e28fd0059fb4d/src/blog.liquid#L21

thekaveman: > `tags` in 11ty are the mechanism by which content is organized into collections.

Ah this makes sense, thanks for the reference.

> https://www.11ty.dev/docs/collections/
>> Take care to note that tags have a singular purpose in Eleventy: to construct collections of content. Some blogging platforms use Tags to refer to a hierarchy of labels for the content

While I understand the reasoning, this is some unfortunate mangling of concepts. :shrug:

thekaveman: > unless i'm missing something, we aren't doing anything in the UI with that information yet, but i'm confident that it'd be straightforward to introduce filtering/search and only syndicate a subset of the blog posts to D+D when we're ready to tackle that.

You are correct! I forgot about these `categories` and that is what I was thinking of with the global `tags`.

src/blog/pretty-atom-feed.xsl
thekaveman: Ah maybe this is only the case in this XSLT transformation of the feed, displayed in browsers?

It seems like the feed as seen by feed readers still contains the full post body, as shown in your validator link: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fdeploy-preview-319--compilerla.netlify.app%2Fblog%2Ffeed.xml

jgravois: unless i've lost the plot, the `feed.xml` still includes the entire body of each blog post and the `.xsl` file is just an additional skin to make it easier to skim the list of posts.

that said, i'll make a note to fire up an actual RSS reader to be absolutely certain this is the case before we merge #302

src/blog/blog.11tydata.js
angela-tran: Yeah, it can get a bit confusing going from the way Jekyll defines and uses "tags", "categories", and "collections" to 11ty's model.

I wanted to answer this part of your question @thekaveman and hopefully I don't add confusion for anyone:

> If we later wanted to apply tags to certain posts, ...  could we still do that on a per-post basis? Would they merge with this global list, or how does that work?

Yes, the tags on the individual post would be merged with the global list. In calitp.org, we have this situation with the Press and Resources posts, where the [directory data file](https://github.com/cal-itp/calitp.org/blob/2da6f010d4f8ea1c3b2ad940a561116d28536c5c/src/resources/resources.json#L2) makes it so all its posts are in that collection, and then each post could have its [own `tags`](https://github.com/cal-itp/calitp.org/blob/2da6f010d4f8ea1c3b2ad940a561116d28536c5c/src/resources/case-study-cal-itp-statewide-fare-guidelines-toolkit.md?plain=1#L7-L8) that we're using for filtering. We create [a separate list](https://github.com/cal-itp/calitp.org/blob/2da6f010d4f8ea1c3b2ad940a561116d28536c5c/src/resources/resources.11tydata.js#L4) that filters out the "collection" tag, so that when we go to render the ~filter UI~ [post's tags](https://github.com/cal-itp/calitp.org/blob/2da6f010d4f8ea1c3b2ad940a561116d28536c5c/src/_includes/article.liquid#L30-L32), we don't include the "collection" tag.

image

jgravois: > this is some unfortunate mangling of concepts. 🤷

i might have just already drank the kool-aid, but personally i think the 11ty model strikes a great balance between being intuitive and flexible. 😆 

individual pages aren't limited to participating in a single collection. beyond appending additional tags, its also easy to create custom collections using JS for filtering.

https://www.11ty.dev/docs/collections-api/

src/blog/pretty-atom-feed.xsl
thekaveman: I'm pretty sure I'm the one that lost the plot. Got confused between the actual feed and the XSLT styled version of the feed that shows up in the browser. Have to say, I've always hated XSLT :sweat_smile: 

This is the kind of change that I'd rather was either documented in a ticket and/or discussed ahead of implementation. It's a fairly big addition / change from where we were before. I think we can merge it but here are some additional thoughts it would have been worth discussing ahead of time:

1. The content that shows up on this page, links out to external sites that we have no affiliation with, etc.
1. The design/branding of this page, differing from Compiler's website and blog as a whole
1. Accessibility considerations
1. Our (potential) need to maintain an XSLT stylesheet going forward

src/blog/pretty-atom-feed.xsl
jgravois: i'm glad you zeroed in on this. turns out the subtitle wasn't rendering either.

fa396ec looks/feels like an improvement to me. let me know what y'all think.

Screenshot 2026-06-09 at 8 50 57 AM

src/blog/pretty-atom-feed.xsl
jgravois: i hear that. if anyone else has reservations about merging as-is i don't mind yanking the .xslt file here and looping back to discuss it independently.

angela-tran: Good discussion here, and I'm ok with merging as-is 👍

Scotchester: Yeah, on second thought, maybe let's pull it for now and revisit later.

jgravois: works for me 20ad781

eleventy.config.js
```diff
@@ -20,4 +20,15 @@ export default async function (eleventyConfig) {
   eleventyConfig.addCollection("jobs", (collectionApi) => {
     return collectionApi.getFilteredByGlob("src/jobs/*.md").sort(byTitle);
   });
+
+  // draft blog posts are rendered locally, but not in production builds
```
lalver1, line 32: With this new preprocessor, can we remove this [comment](https://github.com/compilerla/compiler.la/blob/11ty/.vscode/tasks.json#L23)? It seems like it's the reimplementation we needed.

lalver1 (commented):
I've been reading along and learned a lot, thanks for the comments y'all.
I just have one minor question below, John:

eleventy.config.js
jgravois: great catch @lalver1 

i yanked the entire deprecated task in 516f923

lalver1 (approved):
Looks good! 👍

Issue #201, opened 2024-06-25
Revisit blog listing page


thekaveman:
@laneymangan in Slack: https://compilerla.slack.com/archives/C05L2B0SG06/p1717429574502599

> I noticed that I'm not always able to view all of the past blog posts on our website.

Segacy had originally designed more of a filtering/pagination experience for this page. We should revisit those designs and see what we can implement quickly to make the blog listing page more usable to see older posts as we continue to post more.

A simple initial idea is a "See more" button that just un-hides a large chunk of older posts like how the case studies on the Work page function.

machikoyasuda:
@machikoyasuda Mentioning myself here so I can follow this ticket, and perhaps prioritize it for the next Slush.

thekaveman:
cc @AnthonyRollins @o-ram

PR #321, opened 2026-06-09
chore: fix remaining 11ty todos and discrepancies

jgravois:
resolves #322


.vscode/tasks.json
jgravois, line 1: i often find it handy to generate a one-off build (instead of watching files for changes) so i figured i'd add a task for it to match the one we have in [calitp.org](https://github.com/cal-itp/calitp.org/blob/2da6f010d4f8ea1c3b2ad940a561116d28536c5c/.vscode/tasks.json#L23-L38)

src/_data/feed.json
jgravois, line 1: the new `<meta>` tag in our default.liquid template felt like a good enough excuse to me to centralize the properties of our RSS feed in JSON.

src/_layouts/default.liquid
```diff
@@ -1,40 +1,36 @@
...
-
-    <title>{{ title | default: 'Compiler' }}</title>
+    {% if title_prefix %}
+      {% assign title = title_prefix | append: title %}
+    {% else %}
+      {% assign title = title | default: 'Compiler' %}
+    {% endif %}
 
+    <title>{{ title }}</title>
+    <meta name="og:title" name="title" content="{{ title }}">
```
jgravois, line 11: in the commented code above (and on production) i'd argue that its a logic error to skip injecting the meta tag below when `layout.title_prefix` is defined.

```liquid
<meta name="og:title" name="title" content="{{ page.title }}">
```
but i'm happy to walk back that change if anyone else disagrees.

src/_layouts/default.liquid
```diff
@@ -1,40 +1,36 @@
 <!doctype html>
 <html lang="en">
   <head>
-    {% comment %}
-      TODO: resolve "unexpected 'if page.title'" error
```
jgravois, line 5: i know @angela-tran already knows this, but the crux of the issue here was that front-matter props don't hang off of `page` in 11ty.

src/_layouts/default.liquid
```diff
@@ -1,40 +1,36 @@
...
     {% endif %}
-    {% if page.author %}
-      <meta name="author" content="{{ page.author }}">
+    {% if author %}
+      <meta name="author" content="{{ author }}">
     {% endif %}
     <meta property="og:image" content="/assets/compiler_share.png">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
+    <meta name="generator" content="{{ eleventy.generator }}">
```
jgravois, line 26: this is another change i snuck in for parity's sake. jekyll injects an equivalent tag automatically.

11ty's policy is [opt-in](https://www.11ty.dev/docs/data-eleventy-supplied/#use-with-meta-namegenerator).

src/_layouts/default.liquid
```diff
@@ -1,40 +1,36 @@
...
     <link rel="icon" href="/favicon.ico" sizes="any">
     <link rel="icon" href="/icon.svg" type="image/svg+xml">
     <link rel="apple-touch-icon" href="/apple-touch-icon.png">
 
-    {% comment %}
-      TODO: bring back RSS feed
-      {% feed_meta %}
-    {% endcomment %}
+    <!-- advertise the RSS feed -->
+    <link type="application/atom+xml" rel="alternate" href="{{ blog.atomPath }}" title="{{ blog.title }}">
```
jgravois, line 33: this is something i missed in #307.

src/_layouts/job_post.liquid
```diff
@@ -1,6 +1,5 @@
 ---
 layout: default
-title_prefix: "Jobs with Compiler:"
```
jgravois, line 3: it would be fine to leave this declaration where it is, but it just felt nicer to me to centralize as many shared props in /jobs/jobs.json as possible.

src/_layouts/default.liquid
```diff
@@ -1,40 +1,36 @@
...
-      TODO: resolve "unexpected 'if page.title'" error
-      {% if layout.title_prefix %}
-      {% assign title = layout.title_prefix | append: " " | append: page.title %}
-      {% else if page.title %}
-      {% assign title = page.title %}
-      <meta name="og:title" name="title" content="{{ page.title }}">
-      {% endif %}
-    {% endcomment %}
-
-    <title>{{ title | default: 'Compiler' }}</title>
```
jgravois, line 14: tbh, i'm not sure how/why Jekyll is consistently adding a trailing-space to all the `<title>` tags, but i certainly like it better in this branch without them.

Image

src/_layouts/default.liquid
```diff
@@ -1,40 +1,36 @@
...
+    {% if title_prefix %}
+      {% assign title = title_prefix | append: title %}
```
Scotchester, line 7: This doesn't cause a problem if `title` isn't already defined, does it? It ends up being just the prefix?

Also, we need to reinstate the space that was previously appended between the two, or add the space to the prefix value itself (only one exists: https://github.com/compilerla/compiler.la/blob/10923ca4a480d3641d67805a693f3124ceeb9eac/src/_layouts/job_post.liquid#L3)

src/_layouts/default.liquid
Scotchester: I wouldn't do it in this PR, but it'd be nice to have a visible link or icon in the footer pointing to the feed, too.

src/_data/feed.json
Scotchester: I think it makes sense to centralize the feed config so it can be shared between the `default.liquid` template and `eleventy.config.js`, but I have some quibbles with the naming.

- If these config values are only used by the `feedPlugin`, naming the JSON file `blog.json` seems too generic. Maybe `feed.json`?
- It seems risky/presumptuous to think that we'll never need other variables with such broad names as `base` and `title` in `eleventy.config.js`. Could they just be accessed off of the parent object (`blog` at this time, but again I'd advocate for `feed`) inline? I.e.,
  ```patch
    eleventyConfig.addPlugin(feedPlugin, {
      type: "atom",
  -   outputPath
  +   feed.outputPath,
  ```
  - Note: I did try this locally, following [an example I saw on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with#importing_json_modules_with_the_type_attribute), but it gave me an "Unexpected token '.'" error trying to read `eleventy.config.js`, and I don't know why.
- And back on the first commit of the branch, it's calling for `atomPath` when it should be calling for `outputPath`.

Scotchester (changes requested):

src/_data/feed.json
jgravois: > naming the JSON file `blog.json` seems too generic. Maybe `feed.json`?

no argument here. 👍 

> Could they just be accessed off of the parent object (blog at this time, but again I'd advocate for `feed`) inline?

my thought was that we could/would refactor later as soon as we have other variables that clash, but i'm happy to future-proof it and unwind the shorthand now. 😇 

```js
eleventyConfig.addPlugin(feedPlugin, {
  type: "atom",
  outputPath: feed.outputPath
}
```
https://gomakethings.com/articles/object-property-shorthand-values-with-vanilla-javascript/

src/_layouts/default.liquid
jgravois: > This doesn't cause a problem if title isn't already defined, does it? It ends up being just the prefix?

i'm not opposed to adding another guard, but right now `title_prefix` is defined in jobs.json and 11ty fails to build the site if `title` isn't supplied in an individual job markdown file (because we're using that property to sort).

Screenshot 2026-06-10 at 6 26 09 AM

> Also, we need to reinstate the space that was previously appended between the two

its easy to miss, but i definitely did that when i moved the prop into the jobs.json file.

```jsonc
// jobs.json
{
  "title_prefix": "Jobs with Compiler: "
}
```

src/_data/feed.json
jgravois: done in e3fc474

src/_layouts/default.liquid
Scotchester: > 11ty fails to build the site if `title` isn't supplied in an individual job markdown file

Got it, that's good enough for me, then.

> its easy to miss, but i definitely did that when i moved the prop into the `jobs.json` file.

Wow, sorry. Don't know how I missed that.

src/_data/feed.json
Scotchester: Ahhh, I see now. Thanks!

Scotchester (approved):
LGTM now!

jgravois:
i'll go ahead and merge now, but i'm more than happy to answer post-mortem questions if any come up.

src/_layouts/default.liquid
thekaveman: > it'd be nice to have a visible link or icon in the footer

I don't disagree, but this is another change I'd want to discuss and at least have a Designer weigh in on.

thekaveman: Also note we mention this in https://github.com/compilerla/compiler.la/issues/201#issue-2373705200

thekaveman: @jgravois I'm noting in the preview site that `href` and `title` are empty in this element

jgravois: derp. forgot to update this ref when i addressed @Scotchester's good feedback in e3fc474c231f52f92b0bb7b2f86f50f4f291669b

